PR #2, opened 2016-04-26
[JENKINS-34450] Work around deadlock in core

jglick:
[JENKINS-34450](https://issues.jenkins-ci.org/browse/JENKINS-34450)

While I think the proper fix is in core, it seems like this workaround may avoid the deadlock in the short term. If it seems to work, I can file a core PR, perhaps making `AsynchronousExecution` try to reuse the read/write lock from `Executor`.

@reviewbybees esp. @vivek


svanoort:
:bee: for the changes, assuming it resolves the issue in testing

vivek:
🐝 We can give it a try

jglick:
Released as 2.2-beta-1.

jglick:
Also released as 1.14.1.

PR #3, opened 2016-05-19
If LogRotator runs, we should also ask StashManager to clear all stashes

jglick:
Just like `archive`d files would be deleted at this point, it makes sense to delete `stash`ed files. Normally they are deleted when the build finishes, but plugins can ask to hold onto them for longer.

@reviewbybees esp. @jtnord

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -557,9 +557,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     }
 
     @Override public void deleteArtifacts() throws IOException {
-        synchronized (this) {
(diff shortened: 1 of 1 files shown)


jtnord:
clearing the stashes may well make checkpoints invalid.  Should forcing them to be cleared not call back to the StashManager.StashBehavior so that the user who originally said "do not delete these" can invalidate things that it had that may require them?

jglick:
> clearing the stashes may well make checkpoints invalid

True, but so would deleting artifacts, if you were using `archive` before, and `unarchive` after, a checkpoint—which is also supported (and prior to `stash` was the only option).

This PR is the first step, intended to at least reclaim disk space. The next would probably be to extend `StashBehavior` with a method indicating that the stashes _are_ being cleared, which the checkpoint implementation would implement by deleting associated checkpoints.

jtnord:
:bee:

PR #4, opened 2016-05-27
[JENKINS-33721] - first work on term/kill in side panel.

abayer:
- [x] Add flags to track stop/term/kill calls.
- [x] Modify sidepanel.jelly to provide links to stop/term/kill in appropriate contexts.
- [ ] Get a better icon for the side panel - using same as delete build right now.
- [ ] Tests!

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -127,6 +127,12 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     };
     private transient StreamBuildListener listener;
 
+    private transient boolean stopCalled;
+
+    private transient boolean termCalled;
+
+    private transient boolean killCalled;
+
     /**
      * Flag for whether or not the build has completed somehow.
      * Non-null soon after the build starts or is reloaded from disk.
@@ -288,6 +294,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         if (!isInProgress()) {
             return;
         }
+        termCalled = true;
         final Throwable x = new FlowInterruptedException(Result.ABORTED);
         Futures.addCallback(execution.getCurrentExecutions(/* cf. JENKINS-26148 */true), new FutureCallback<List<StepExecution>>() {
             @Override public void onSuccess(List<StepExecution> l) {
@@ -316,6 +323,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         if (!isBuilding()) {
             return;
         }
+        killCalled = true;
         if (listener != null) {
             listener.getLogger().println("Hard kill!");
         }
@@ -324,6 +332,18 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         finish(Result.ABORTED, suddenDeath);
         executionPromise.setException(suddenDeath);
         // TODO CpsFlowExecution.onProgramEnd does some cleanup which we cannot access here; perhaps need a FlowExecution.halt(Throwable) API?
+    }
+
+    public boolean hasStopBeenCalled() {
+        return stopCalled;
+    }
+
+    public boolean hasTermBeenCalled() {
+        return termCalled;
+    }
+
+    public boolean hasKillBeenCalled() {
+        return killCalled;
     }
 
     @GuardedBy("completed")
@@ -634,6 +654,9 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
 
     @RequirePOST
     public synchronized HttpResponse doStop() {
+        if (hasPermission(Item.CANCEL)) {
+            stopCalled = true;
+        }
         Executor e = getOneOffExecutor();
         if (e != null) {
             return e.doStop();
--- a/src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
@@ -34,6 +34,37 @@
             <p:console-link/>
             <l:task icon="images/24x24/notepad.png" href="${buildUrl.baseUrl}/configure" title="${h.hasPermission(it,it.UPDATE)?'%Edit Build Information':'%View Build Information'}"/>
             <st:include page="delete.jelly"/>
+            <j:if test="${it.isBuilding() and h.hasPermission(it,it.CANCEL)}">
+                <j:choose>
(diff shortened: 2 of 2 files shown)


abayer:
Note - I've tested this by hand and stop shows up right, but I'm not sure how to get to the term/kill cases. I'm also not sure how to write unit tests for this in the first place. =)

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
```diff
@@ -34,6 +34,37 @@
             <p:console-link/>
             <l:task icon="images/24x24/notepad.png" href="${buildUrl.baseUrl}/configure" title="${h.hasPermission(it,it.UPDATE)?'%Edit Build Information':'%View Build Information'}"/>
             <st:include page="delete.jelly"/>
+            <j:if test="${it.isBuilding() and h.hasPermission(it,it.CANCEL)}">
+                <j:choose>
+                    <j:when test="${it.hasStopBeenCalled() == false}">
```
jglick: `!it.hasStopBeenCalled()`?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -634,6 +654,9 @@ public boolean hasntStartedYet() {
 
     @RequirePOST
     public synchronized HttpResponse doStop() {
+        if (hasPermission(Item.CANCEL)) {
+            stopCalled = true;
```
jglick: Possibly better to move this [here](https://github.com/jenkinsci/workflow-job-plugin/blob/ba83d8f854b6befcb64c1b824ecdef207658b369/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L238).

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
```diff
@@ -34,6 +34,37 @@
...
+                        <l:task icon="images/24x24/edit-delete.png"
+                                href="${buildUrl.baseUrl}/stop"
+                                title="${%Stop Build}"
```
jglick: Do we really want to display this? It would be inconsistent with the UI for freestyle builds; there is already a fairly prominent stop button not just on the `console` page but in the build history widget.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
```diff
@@ -34,6 +34,37 @@
...
+                    <j:when test="${it.hasTermBeenCalled() == false}">
+                        <l:task icon="images/24x24/edit-delete.png"
+                                href="${buildUrl.baseUrl}/term"
+                                title="${%Abort Running Steps}"
+                                contextMenu="false"
+                                post="true"
+                                requiresConfirmation="true"
+                                confirmationMessage="Really forcibly abort running steps?"/>
+                    </j:when>
+                    <j:when test="${it.hasKillBeenCalled() == false}">
```
jglick: No, we do _not_ want to display the `kill` link instantly after `term` has been called. Similarly, we do not want to display the `term` link instantly after `stop` has been called. The 15s delay is deliberate, to give the build a chance to stop cleanly.

jglick:
> I'm not sure how to get to the term/kill cases

Easiest example:

```groovy
while (true) {
  try {
    sleep 999
  } catch (Exception belaLugosiIsDead) {
    echo 'undead undead undead'
  }
}
```

> not sure how to write unit tests for this in the first place

`WorkflowRunRestartTest.hardKill`

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
abayer: Weirdly, that didn't work. I'll poke at it again, but yeah...

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
abayer: Ok.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
abayer: I'm open to alternative locations for this, but it definitely has to be more than just in the log. I suppose we could just transform the stop button's actions on the fly?

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
jglick: Also the current UI allows you to try `stop` multiple times without using `term`, or try `term` multiple times without using `kill`. Each `stop` will send a polite signal to the running step(s); even when the first signal did not suffice, a second might. Each `term` will immediately halt the running step(s); even if stopping that step did not halt the build, stopping a _later_ step might. For example:

```groovy
catchError {
  someHangingStepThatCannotBeInterrupted()
}
mail …
```

If you `stop`, nothing will happen, and after 15s you get the option to `term`. Say you take it, killing the hanging step. Now the build proceeds to the `mail` step. Say this in turn takes too long, perhaps trying to contact a unresponsive SMTP server. A simple `stop` should suffice to kill `mail` politely. You do not need `term` or `kill`.

Basically we want to avoid escalating unless and until it is really necessary.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
jglick: > it definitely has to be more than just in the log

Not following. The option to `stop` was already in multiple places, just as for freestyle builds.

> we could just transform the stop button's actions

As noted below, I would rather keep `term` and `kill` as separate, explicit options, because they have more side effects than `stop`.

abayer: Ok, then we need to pop up a button somewhere for `term` and `kill` when appropriate - where would you want that? In the sidebar, like here, but no `stop` button?

jglick: I guess so. Or add the `stop` link only once it has already been used.

My main concern here is the UI inconsistency with freestyle builds in the normal case that `stop` suffices. There are probably a number of viable UI choices, including changing the UI of freestyle builds to add an explicit _Stop_ link. I do not have a strong opinion.

abayer:
Abandoning this. May come back to it later in a new form.

PR #5, opened 2016-06-09
[JENKINS-26130] Way to print progress from pickles

jglick:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/5 and https://github.com/jenkinsci/workflow-support-plugin/pull/4.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -754,6 +754,16 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         @Override public String getUrl() throws IOException {
             return run().getUrl();
         }
+        @Override public TaskListener getListener() throws IOException {
+            StreamBuildListener l = run().listener;
+            if (l != null) {
+                return l;
+            } else {
+                // TODO this seems to happen at least once during resume; can we get the listener ready earlier?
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -754,6 +754,16 @@ private String key() {
         @Override public String getUrl() throws IOException {
             return run().getUrl();
         }
+        @Override public TaskListener getListener() throws IOException {
+            StreamBuildListener l = run().listener;
+            if (l != null) {
+                return l;
+            } else {
+                // Seems to happen at least once during resume, but anyway TryRepeatedly will call this method again, rather than caching the result.
+                LOGGER.log(Level.FINE, "No listener yet for {0}", this);
```
oleg-nenashev: Maybe FINEST since it's a known issue? I can only imagine how much spam it may generate on the FINE level

oleg-nenashev:
🐝 excepting the log level

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: In practice I have only seen it printed once per build; the `listener` is set during `onLoad`, just apparently slightly later than when the pickles start dehydrating.

oleg-nenashev:
@reviewbybees done

PR #8, opened 2016-07-12
[JENKINS-34547] Switch WorkflowJob.concurrentBuild to JobProperty

abayer:
[JENKINS-34547](https://issues.jenkins-ci.org/browse/JENKINS-34547)

- [x] First attempt at implementation.
- [x] Tests written - see https://github.com/jenkinsci/workflow-multibranch-plugin/pull/19
- [x] Verify backwards compatibility.
- [x] Verify UI works as expected.

cc @reviewbybees most especially @jglick


abayer:
Hmm - looks like `JobPropertyStep` may need to be moved into here for test purposes, at least.

abayer:
Or the test needs to go in `workflow-multibranch-plugin.git`...

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
```diff
@@ -0,0 +1,68 @@
...
+    public synchronized Boolean getAllowConcurrentBuilds() {
+        return allowConcurrentBuilds;
+    }
+
+    synchronized void setAllowConcurrentBuilds(boolean b) {
+        this.allowConcurrentBuilds = b ? null : false;
+    }
+
+    @Extension
+    @Symbol("concurrentBuildProperty")
```
jglick: :bug: do not include `property` in the symbol name. It is implicit in the extension point.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
```diff
@@ -0,0 +1,68 @@
...
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
+
+import javax.annotation.CheckForNull;
+
+/**
+ * {@link WorkflowJobProperty} for setting whether a job should allow concurrent builds.
+ */
+@ExportedBean
+public class WorkflowConcurrentBuildJobProperty extends WorkflowJobProperty {
```
jglick: :ant: just extend `JobProperty`.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
```diff
@@ -0,0 +1,68 @@
...
+    public WorkflowConcurrentBuildJobProperty(boolean b) {
+        this.allowConcurrentBuilds = b ? null : false;
+    }
+
+    @Exported
+    public synchronized Boolean getAllowConcurrentBuilds() {
+        return allowConcurrentBuilds;
+    }
+
+    synchronized void setAllowConcurrentBuilds(boolean b) {
```
jglick: :ant: unused, delete

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
```diff
@@ -0,0 +1,68 @@
...
+
+import javax.annotation.CheckForNull;
+
+/**
+ * {@link WorkflowJobProperty} for setting whether a job should allow concurrent builds.
+ */
+@ExportedBean
+public class WorkflowConcurrentBuildJobProperty extends WorkflowJobProperty {
+    /** Defaults to true */
+    private @CheckForNull Boolean allowConcurrentBuilds;
```
jglick: :bug: use `boolean`. The absence of this property can be construed as an implicit true.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowJob/configure-entries.jelly
```diff
@@ -25,7 +25,6 @@
   -->
 
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:p="/lib/hudson/project">
-    <p:config-concurrentBuild/>
```
jglick: Need to replace with the `BuildAuthorizationToken` configuration snippet.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
jglick: Or, if you want to _remove_ this property when allowing concurrent builds, then just delete the field altogether, and extend `OptionalJobProperty`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -106,7 +108,7 @@
     @SuppressWarnings("deprecation")
     private hudson.model.BuildAuthorizationToken authToken;
     private transient LazyBuildMixIn<WorkflowJob,WorkflowRun> buildMixIn;
-    /** defaults to true */
+    /** @deprecated replaced by {@link WorkflowConcurrentBuildJobProperty} defaults to true */
```
jglick: delete the original Javadoc

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
jglick: Or conversely, you could have the presence of this property mean to allow concurrent builds, the absense to not.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
abayer: Okiedokie - I was wary of just going with "concurrentBuild" since this is Pipeline-specific...

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
abayer: Gonna go with the `OptionalJobProperty` approach, I think.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
abayer: Ok.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -339,12 +340,24 @@ public String getShortDescription() {
...
+        if (p != null) {
+            return p.getAllowConcurrentBuilds();
+        } else {
+            /* settings compatibility */
+            return !Boolean.FALSE.equals(concurrentBuild);
+        }
     }
     
     public void setConcurrentBuild(boolean b) throws IOException {
-        concurrentBuild = b ? null : false;
```
jglick: I think you now need to unconditionally set `concurrentBuild` to null. And deprecate this method.

Either way, you should start off by writing the `@LocalData`-based tests which define what you expect to happen w.r.t. settings compatibility.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/WorkflowConcurrentBuildJobProperty.java
abayer: The current default case is to allow concurrent builds, so I think we'd want to go with `OptionalJobProperty`...

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Will do!

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Ok, will do.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -138,6 +140,9 @@ public WorkflowJob(ItemGroup parent, String name) {
         for (Trigger t : triggers) {
             t.start(this, Items.currentlyUpdatingByXml());
         }
+        if (concurrentBuild != null) {
```
abayer: Not happy about this, but `setConcurrentBuild(boolean)` alone wasn't enough to get the migration working.

jglick:
Need to use `configRoundTrip` to see that the property can be saved & redisplayed on a non-multibranch project.

abayer:
@jglick - clarifying - you mean just a `JenkinsRule.configRoundTrip(job)`?

abayer:
So the `OptionalJobProperty` approach has actually broken this logic pretty hard - need to do some reworking.

abayer:
There's a nasty little contradiction - if we want just a checkbox in the UI, we need `OptionalJobProperty`, but if we want to have concurrent builds enabled by default with a way to disable them from the `properties` step, `OptionalJobProperty` doesn't work. Grr.

jglick:
> you mean just a `JenkinsRule.configRoundTrip(job)`

Yes.

> if we want to have concurrent builds enabled by default

No strong opinion. Originally this was the only option for Pipeline. But for freestyle projects, the default has always been nonconcurrent, with the option for concurrent added later. Certainly we need to preserve the behavior of existing jobs.

> with a way to disable them from the `properties` step, `OptionalJobProperty` doesn't work

Why is that?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -138,6 +140,9 @@ public WorkflowJob(ItemGroup parent, String name) {
         for (Trigger t : triggers) {
             t.start(this, Items.currentlyUpdatingByXml());
         }
+        if (concurrentBuild != null && !concurrentBuild) {
```
jglick: Do you not want to call `setConcurrentBuild` even if it is false, just in case?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -339,12 +343,27 @@ public String getShortDescription() {
 
     @Exported
     @Override public boolean isConcurrentBuild() {
-        return !Boolean.FALSE.equals(concurrentBuild);
+        DisableConcurrentBuildsJobProperty p = getProperty(DisableConcurrentBuildsJobProperty.class);
+        if (p != null) {
+            return false;
+        } else {
+            /* settings compatibility */
+            return !Boolean.FALSE.equals(concurrentBuild);
```
jglick: Can just become

```java
return true;
```

now, since `onLoad` would already have cleared the field.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -339,12 +343,27 @@ public String getShortDescription() {
...
+
     public void setConcurrentBuild(boolean b) throws IOException {
-        concurrentBuild = b ? null : false;
-        save();
+        concurrentBuild = null;
+        BulkChange bc = new BulkChange(this);
+        try {
+            removeProperty(DisableConcurrentBuildsJobProperty.class);
+            if (!b) {
+                addProperty(new DisableConcurrentBuildsJobProperty());
```
jglick: Could be optimized to skip a `save` in case the setting does not change.

src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
```diff
@@ -276,4 +277,21 @@ public void failedToStartRun() throws Exception {
         assertEquals(Collections.emptyList(), iba.getCauses());
     }
 
+    @Issue("JENKINS-34547")
+    @LocalData
+    @Test public void concurrentBuildsMigrationOnByDefault() throws Exception {
```
jglick: :ant: the code in question is in `WorkflowJob`, so it would make more sense for this to be in a `WorkflowJobTest`. Or in a `DisableConcurrentBuildsJobPropertyTest`.

jglick:
:bee:

jglick:
>> if we want to have concurrent builds enabled by default
> No strong opinion.

Though I tend to prefer concurrent by default, since to my way of thinking the serial option is a special behavior, akin to a `QueueTaskDispatcher` which blocks builds from running under some conditions ([example](https://github.com/jenkinsci/branch-api-plugin/blob/1626b7db03cad1a83b163491aff5f372384735b7/src/main/java/jenkins/branch/RateLimitBranchProperty.java#L379-L398))—in this case, just that another build of the same job is currently running.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Yeah, you're right there. That was an artifact of earlier efforts. Fixing!

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Even simpler - 
```
return getProperty(DisableConcurrentBuildsJobProperty.class) == null;
```

=)

src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
abayer: Yup, moved around.

abayer:
So given that the default behavior has been to allow concurrent builds, I think that *has* to stay the same. Therefore, switching the UI from a checked-by-default checkbox for enabling concurrent builds to an unchecked-by-default checkbox for disabling concurrent builds seems eminently reasonable for me.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Adding such an optimization with a few line comment explaining what exactly the logic is for the optimization, since it's a bit weird at first glance. =)

jglick:
> given that the default behavior has been to allow concurrent builds, I think that has to stay the same

Well…not for jobs saved since the `concurrentBuild` field was added. But for really old jobs last saved before there was even the option, yes.

> switching the UI from a checked-by-default checkbox for enabling concurrent builds to an unchecked-by-default checkbox for disabling concurrent builds seems eminently reasonable for me

I am not opposed to that necessarily, though users might be confused that the UI is opposite that of freestyle projects. If we want to keep a consistent UI then we cannot use `OptionalJobProperty`, I am afraid.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -343,26 +343,28 @@ public String getShortDescription() {
...
 
     public void setConcurrentBuild(boolean b) throws IOException {
         concurrentBuild = null;
-        BulkChange bc = new BulkChange(this);
-        try {
-            removeProperty(DisableConcurrentBuildsJobProperty.class);
-            if (!b) {
-                addProperty(new DisableConcurrentBuildsJobProperty());
+
+        boolean propertyExists = getProperty(DisableConcurrentBuildsJobProperty.class) == null;
```
jglick: :ant: the variable name is the opposite of the meaning.

abayer: Also the comment was the inverse of what was actually being done. Whoops!

abayer:
Yeah, the UI inconsistency is the only way I see to have just a checkbox with no further config in the UI, where it can also be disabled via the `properties` step. Not to mention there's a challenge in getting an `OptionalJobProperty` to be selected by default!

jglick:
> the UI inconsistency is the only way I see to have just a checkbox with no further config in the UI, where it can also be disabled via the `properties` step

Well the alternative is a non-`Optional` `JobProperty` whose `config` view has a checkbox in it, and from `Jenkinsfile` you would need write something like

```groovy
properties([concurrentBuilds(false)])
```

jglick:
re:bee:

abayer:
Yeah, I considered the traditional JobProperty approach but dislike that UX more than changing the behavior.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -339,12 +343,29 @@ public String getShortDescription() {
...
+        if (propertyExists == b) {
+            BulkChange bc = new BulkChange(this);
+            try {
+                removeProperty(DisableConcurrentBuildsJobProperty.class);
+                if (!b) {
+                    addProperty(new DisableConcurrentBuildsJobProperty());
+                }
+                bc.commit();
+            } finally {
+                bc.abort();
```
svanoort: Why are we aborting after committing when successful?

abayer: That was lifted from somewhere - lemme find it. I think that the `abort` ends up doing nothing post-`commit`....

abayer: Ah, this was the paradigm used before https://github.com/jenkinsci/jenkins/commit/5b4dbb3970b7e559e32aee6dbd159e66bf7a68a8, and since we're not basing on HEAD of core, we still need to do things the old way.

svanoort: I mean yeah it shouldn't, but it's odd to me to do it there rather than an abort and re-throw on error.

abayer: I'm assuming that the paradigm is saying "Eh, if it goes wrong, just don't do it and move on"?

abayer: And `abort` just returns immediately if the change is already complete, so there's no harm in calling it after success.

svanoort: I mean yeah it's not a bug, but it's just a weird approach and a bit unexpected -- in the event that commit/abort become more involved/complex (ex: pluggable storage or similar), this codepath could have unexpected side effects.

abayer: *shrug*

abayer: I'd say the real issue isn't the code path - it's the name `abort`, since what it really should be is something like `cleanup` since it only aborts if the change is incomplete. Anyway, after that commit I linked above, it's irrelevant.

svanoort: Although... the alternative is pretty ugly (catch Exception, try to abort, then finally rethrow exception?).

svanoort: Meh. 🐝  and some grumping about the BulkChange APIs from me.

abayer:
Yay, @reviewbybees done

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
jglick: The code is correct here.

PR #9, opened 2016-07-13
JENKINS-34005 - Convert WorkflowJob.triggers to a JobProperty

abayer:
[JENKINS-34005](https://issues.jenkins-ci.org/browse/JENKINS-34005)

- [x] Add new `WorkflowTriggersJobProperty`
- [x] Testing adding/removing/setting triggers
- [x] Testing migration from old style.
- [x] Verify UI behavior including config round tripping.
- [x] Verify comparison of behavior with `TriggerStartTest` for `AbstractProject` in core.
- [x] `properties` step test over in `workflow-multibranch-plugin` (see https://github.com/jenkinsci/workflow-multibranch-plugin/pull/20)

cc @reviewbybees esp @jglick


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -178,17 +182,22 @@ public void setDefinition(FlowDefinition definition) {
         JSONObject json = req.getSubmittedForm();
         definition = req.bindJSON(FlowDefinition.class, json.getJSONObject("definition"));
         authToken = hudson.model.BuildAuthorizationToken.create(req);
-        for (Trigger t : triggers) {
-            t.stop();
+        WorkflowTriggersJobProperty triggerProp = getTriggersJobProperty();
+        if (triggerProp != null) {
+            for (Trigger t : triggerProp.getTriggers()) {
+                t.stop();
```
jglick: Seems antimodular. This should be done in `WorkflowTriggersJobProperty` itself.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -178,17 +182,22 @@ public void setDefinition(FlowDefinition definition) {
         JSONObject json = req.getSubmittedForm();
         definition = req.bindJSON(FlowDefinition.class, json.getJSONObject("definition"));
         authToken = hudson.model.BuildAuthorizationToken.create(req);
-        for (Trigger t : triggers) {
-            t.stop();
+        PipelineTriggersJobProperty triggerProp = getTriggersJobProperty();
+        if (triggerProp != null) {
+            for (Trigger t : triggerProp.getTriggers()) {
+                t.stop();
```
jglick: Antimodular. Do this in `WorkflowTriggersJobProperty`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -132,13 +135,14 @@ public WorkflowJob(ItemGroup parent, String name) {
...
-            triggers.setOwner(this);
+        if (triggers != null) {
+            setTriggers(triggers.toList());
         }
-        for (Trigger t : triggers) {
-            t.start(this, Items.currentlyUpdatingByXml());
+        // Since we would have already started them in setTriggers...
+        else {
+            for (Trigger t : getTriggersJobProperty().getTriggers()) {
+                t.start(this, Items.currentlyUpdatingByXml());
```
jglick: Do this in `WorkflowTriggersJobProperty.setJob`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -437,24 +446,83 @@ public void setConcurrentBuild(boolean b) throws IOException {
...
+            bc.commit();
+
+            if (originalProp != null) {
+                for (Trigger t : originalProp.getTriggers()) {
+                    t.stop();
+                }
+            }
+
+            for (Trigger t2 : triggerProp.getTriggers()) {
+                t2.start(this, true);
```
jglick: Could more simply edit the existing property and `reconfigure` it.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -437,24 +446,83 @@ public void setConcurrentBuild(boolean b) throws IOException {
...
+                originalProp = new PipelineTriggersJobProperty(Collections.<Trigger<?>>singletonList(trigger));
+            }
+
+            addProperty(originalProp);
+            bc.commit();
+
+            if (old != null) {
+                old.stop();
+            }
+            trigger.start(this, true);
```
jglick: Ditto.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -483,7 +551,8 @@ public void replaceAction(Action a) {
     }
 
     @Override public SCMTrigger getSCMTrigger() {
-        return triggers.get(SCMTrigger.class);
+        // TODO: This seems fairly weird to me and I'm not 100% sure why it works.
+        return (SCMTrigger) getTriggers().get(getDescriptorByName("SCMTrigger"));
```
jglick: Do not do it this way.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Yeah, realizing that.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: `setJob`? Do you mean `setOwner`?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Yeah, gonna do the latter.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: So what way *should* I do it?

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @CheckForNull
+    @Override
+    public PipelineTriggersJobProperty reconfigure(@Nonnull StaplerRequest req, @CheckForNull JSONObject form) throws Descriptor.FormException {
+        PipelineTriggersJobProperty thisProp;
+
+        // TODO: See if we actually need to check this or can assume it's always non-null
+        if (form != null) {
+            thisProp = (PipelineTriggersJobProperty)getDescriptor().newInstance(req, form);
```
abayer: So it turns out this straight-up doesn't work. Yay! I've verified that the `form` content is correct (with my current `config.jelly`), but by the time it gets to the `PipelineTriggersJobProperty` constructor, it's getting a `null` list of `Trigger`s. So something in the binding is not working with my approach. Grr! Am I forgetting something obvious?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
jglick: Whatever it is.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
jglick: Just search the list for `instanceof SCMTrigger`? There is a helper function somewhere in `Util` IIRC.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
jglick: Do not call `newInstance` I guess. Use `req.bindJSON(this, form)` or something along those lines. Or do something a bit lower-level, pulling the list out of the `form` and using `DescriptorList` methods IIRC to data-bind it to a list.

abayer: Okiedokie - will try those when I get home.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -437,24 +446,83 @@ public void setConcurrentBuild(boolean b) throws IOException {
...
 
     @SuppressWarnings("deprecation")
     @Override public List<Action> getActions() {
         List<Action> actions = new ArrayList<Action>(super.getActions());
-        for (Trigger<?> trigger : triggers) {
-            actions.addAll(trigger.getProjectActions());
+        PipelineTriggersJobProperty triggerProp = getTriggersJobProperty();
+        if (triggerProp != null) {
+            for (Trigger<?> trigger : triggerProp.getTriggers()) {
+                actions.addAll(trigger.getProjectActions());
```
abayer: Would that mean changing to `super.getAllActions()`, actually?

abayer:
So as it turns out, calling `startTriggers()` in `PipelineTriggersJobProperty.setOwner` causes...chaos. Way too many calls to `startTriggers()` with confusing variants on the `newInstance` flag. The only way I can replicate behavior from `TriggerStartTest` in core is to remove the call from `setOwner` and put it back in `WorkflowJob.onLoad`, so, well, I'm doing that.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -128,21 +131,19 @@ public WorkflowJob(ItemGroup parent, String name) {
...
         buildMixIn.onLoad(parent, name);
-        if (triggers == null) {
-            triggers = new DescribableList<Trigger<?>,TriggerDescriptor>(this);
-        } else {
-            triggers.setOwner(this);
-        }
-        for (Trigger t : triggers) {
-            t.start(this, Items.currentlyUpdatingByXml());
+        if (triggers != null && !triggers.isEmpty()) {
+            setTriggers(triggers.toList());
```
jglick: :bug: this is going to call `startTriggers(true)` whereas it should usually be false (unless we are `currentlyUpdatingByXml`).

jglick: Only affects the one-time load after upgrading this plugin, but still—for some plugins the value of `newInstance` may be important.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -437,26 +432,59 @@ public void setConcurrentBuild(boolean b) throws IOException {
...
-        return triggers.toMap();
+        return getTriggersJobProperty().getTriggersMap();
     }
 
-    public void addTrigger(Trigger trigger) {
-        Trigger old = triggers.get(trigger.getDescriptor());
-        if (old != null) {
-            old.stop();
-            triggers.remove(old);
+    public PipelineTriggersJobProperty getTriggersJobProperty() {
```
jglick: Could be `private` I think.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -0,0 +1,177 @@
...
+        for (Trigger trigger : triggers) {
+            if (td.equals(trigger.getDescriptor())) {
+                return trigger;
+            }
+        }
+
+        return null;
+    }
+
+    public void stopTrigggers() {
```
jglick: :bug: typo (would be an :ant: if made package-private)

src/main/resources/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty/config.jelly
```diff
@@ -0,0 +1,39 @@
...
+-->
+
+<!--
+  Configure build triggers, applicable to ParameterizedJobMixIn.ParameterizedJob
+-->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <!-- build triggers config pane -->
+    <j:invokeStatic var="triggers" className="hudson.triggers.Trigger" method="for_">
+        <j:arg value="${it}" type="hudson.model.Item" />
```
jglick: I tend to prefer encapsulating such logic in a method in the `DescriptorImpl` so the Jelly view stays short and simple.

abayer: Makes sense - this was copy-pasted from `config-triggers.jelly`, or I would have done something similar already. =)

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Ok, will change to use `currentlyUpdatingByXml` here.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Done.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
abayer: Fixed.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Actually, no, it can't - needed in the `TransientActionFactory` for `PipelineTriggersJobProperty`.

src/main/resources/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty/config.jelly
abayer: Actually, off the top of my head (and admittedly at the end of the day, so my head may not be all there), I'm not sure of an easy way to encapsulate it in `WorkflowJob.DescriptorImpl`, so I'm gonna leave it as is at least for tonight.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: More accurately, I changed `setTriggers(list)`'s call to `startTriggers` to `startTriggers(Items.currentlyUpdatingByXml)`.

jglick: But that will be wrong for calls from anywhere but `onLoad`, where we expect true, and `currentlyUpdatingByXml` will likely be false.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -437,26 +438,59 @@ public void setConcurrentBuild(boolean b) throws IOException {
...
+            removeProperty(PipelineTriggersJobProperty.class);
+
+            PipelineTriggersJobProperty triggerProp = new PipelineTriggersJobProperty(inputTriggers);
+
+            addProperty(triggerProp);
+            bc.commit();
+
+            originalProp.stopTriggers();
+
+            triggerProp.startTriggers(Items.currentlyUpdatingByXml());
```
jglick: Maybe this should just be deleted now, since `onLoad` is calling it already?

jglick: Or rather, could be moved into `submit`.

src/test/java/org/jenkinsci/plugins/workflow/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -0,0 +1,177 @@
...
+        assertEquals(1, roundTripWithTrigger.getTriggers().size());
+        List<Trigger<?>> modTriggers = new ArrayList<>(roundTripWithTrigger.getTriggers().values());
+
+        assertEquals(MockTrigger.class, modTriggers.get(0).getClass());
+    }
+
+    public static class MockTrigger extends Trigger<Item> {
+
+        public transient List<Boolean> calls = new ArrayList<Boolean>();
+        public transient boolean isStarted = false;
```
jglick: Might be more convincing for these to be replaced with a single

```java
/** elements are true or false for {@code start}, null for {@code stop} */
public static List<Boolean> startsAndStops = new ArrayList<>();
```

so that you can demonstrate that the correct number and kind of calls are made on both old and new instances: that there are not, for example, temporary `Trigger`s constructed, `start`ed, then discarded without being `stop`ped.

This would be closer to the real behavior of `Trigger`s overriding these methods, which tend to do things like register webhooks etc. with some global effect, where you want to ensure that calls are properly paired.

src/test/java/org/jenkinsci/plugins/workflow/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -0,0 +1,177 @@
...
+        PipelineTriggersJobProperty triggerProp = p.getProperty(PipelineTriggersJobProperty.class);
+        assertNotNull(triggerProp);
+        assertEquals(1, triggerProp.getTriggers().size());
+        assertEquals(1, p.getTriggers().size());
+
+        Trigger fromProp = triggerProp.getTriggers().get(0);
+        assertEquals(TimerTrigger.class, fromProp.getClass());
+
+        Trigger fromJob = p.getTriggers().get(fromProp.getDescriptor());
+        assertEquals(fromProp, fromJob);
```
jglick: This also needs to be measuring the `start`/`stop` calls. I suspect that in your current code, `start` is called twice.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: `submit` is already (effectively) calling `startTriggers` via `PipelineTriggersJobProperty.reconfigure`.

src/test/java/org/jenkinsci/plugins/workflow/properties/PipelineTriggersJobPropertyTest.java
abayer: Good idea. Gonna do that and switch to deploying a test jar here for reuse over in https://github.com/jenkinsci/workflow-multibranch-plugin/pull/20

src/test/java/org/jenkinsci/plugins/workflow/properties/PipelineTriggersJobPropertyTest.java
abayer: Will check - it's definitely only getting called once in `loadCallsStartFalse` and `submitCallsStartTrue` or they'd be failing on the `t.calls.toString()` assert.

src/test/java/org/jenkinsci/plugins/workflow/properties/PipelineTriggersJobPropertyTest.java
abayer: aaaand good call - there's some weirdness with `start`/`stop`!

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -178,20 +179,21 @@ public void setDefinition(FlowDefinition definition) {
...
-        }
+
         if (req.getParameter("hasCustomQuietPeriod") != null) {
             quietPeriod = Integer.parseInt(req.getParameter("quiet_period"));
         } else {
             quietPeriod = null;
         }
-        triggers.rebuild(req, json, Trigger.for_(this));
-        for (Trigger t : triggers) {
-            t.start(this, true);
```
jglick: Now being handled in `reconfigure` IIUC.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -0,0 +1,176 @@
...
+        PipelineTriggersJobProperty thisProp = new PipelineTriggersJobProperty(new ArrayList<>(trigList.toList()));
+        thisProp.setOwner(owner);
+
+        thisProp.startTriggers(true);
+
+        return thisProp;
+    }
+
+    @Extension
+    @Symbol("pipelineTriggers")
```
jglick: @kohsuke it is a pity that the symbol must be unique amongst all `JobProperty`s, even though `isApplicable` (inherited and using the `WorkflowJob` type parameter) means that for a given job or job type just `triggers` would suffice. But I guess fixing that could be complicated: `resolveClass` in `structs` would need to be supplied with a filter predicate to apply to candidate extensions.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
+    private List<Trigger<?>> triggers = new ArrayList<>();
+
+    @DataBoundConstructor
+    public PipelineTriggersJobProperty(List<Trigger<?>> triggers) {
+        // Defensive handling of when we get called via {@code Descriptor.newInstance} with no form data.
```
jglick: Unfortunately Jenkins’ crazy form binding code often uses different JSON data types (null vs. object vs. array) for lists with zero, one, or multiple elements, resp.

src/test/java/org/jenkinsci/plugins/workflow/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -0,0 +1,182 @@
...
+        MockTrigger.startsAndStops = new ArrayList<>();
+    }
+
+    @Test
+    public void loadCallsStartFalse() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        MockTrigger t = new MockTrigger();
+        p.addTrigger(t);
+        p.save();
+        p = (WorkflowJob) Items.load(p.getParent(), p.getRootDir());
```
jglick: :ant: More realistic simulations would include:

* use `@LocalData` with a `config.xml` representing an existing trigger
* add as above, but use `RestartableJenkinsRule` and do this test after the restart
* use `AbstractItem.doReload`

src/test/java/org/jenkinsci/plugins/workflow/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -0,0 +1,182 @@
...
+    @Test
+    public void triggerMigration() throws Exception {
+        WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+        assertNotNull(p);
+
+        PipelineTriggersJobProperty triggerProp = p.getProperty(PipelineTriggersJobProperty.class);
+        assertNotNull(triggerProp);
+
+        assertEquals(2, triggerProp.getTriggers().size());
+        assertEquals(2, p.getTriggers().size());
```
jglick: Seems like it would be simpler to just assert that these two lists are equal, then check each element once?

src/test/java/org/jenkinsci/plugins/workflow/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -0,0 +1,182 @@
...
+        List<Trigger<?>> origTriggers = new ArrayList<>(withTriggerCase.getTriggers().values());
+
+        assertEquals(MockTrigger.class, origTriggers.get(0).getClass());
+
+        WorkflowJob roundTripWithTrigger = r.configRoundtrip(withTriggerCase);
+
+        assertEquals(1, roundTripWithTrigger.getTriggers().size());
+        List<Trigger<?>> modTriggers = new ArrayList<>(roundTripWithTrigger.getTriggers().values());
+
+        assertEquals(MockTrigger.class, modTriggers.get(0).getClass());
```
jglick: :ant: would be nice to include some trigger with configuration, such as `TimerTrigger`, in the list and verify that this configuration got correctly round-tripped by `f:descriptorList`.

jglick:
There are a couple of places where I think the tests could be improved, but :bee: and congratulations on your patience with this slog!

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Yes.

abayer:
@jglick So given said slog, I'd say let's merge it as is and then I'll do a followup to address the :ant:s?

jglick:
Sure.

KostyaSha:
So since this version my triggers are broken.

KostyaSha:
Why design is screwed now for such wide installed plugin?

jglick:
If there seems to be a regression, please file details to reproduce in a minimal demo trigger and probably @abayer will take a look.

KostyaSha:
@jglick yes, i also reproducing. Just don't understand why such hacking is needed for fresh plugin. Multibranch may be designed in any way as it still new plugin. 

https://issues.jenkins-ci.org/browse/JENKINS-38454

PR #10, opened 2016-07-14
Actually round-trip the disabled concurrent builds case.

abayer:
While working on something else, I noticed that the default case and
disabled case sections in
`DisableConcurrentBuildsJobPropertyTest#configRoundTrip` were
identical. Which kinda defeated the point. So here we go - actually
roundtrip the disabled case.

cc @reviewbybees, esp @jglick


jglick:
:bee:

PR #14, opened 2016-08-03
Overdue TODO comments, -source 7 updates, etc.

jglick:
Just stuff that has collected after baseline updates and the like.

@reviewbybees


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -319,29 +319,13 @@ public void setQuietPeriod(Integer seconds) throws IOException {
...
-            return new BecauseOfBuildInProgress(getLastBuild());
+            WorkflowRun lastBuild = getLastBuild();
+            if (lastBuild != null) {
+                return new BlockedBecauseOfBuildInProgress(lastBuild);
+            } // else race condition, cf. AbstractProject
         }
         return null;
     }
-    // TODO use BlockedBecauseOfBuildInProgress in 1.624 (and remove Messages.properties in resources)
-    public static class BecauseOfBuildInProgress extends CauseOfBlockage {
```
oleg-nenashev: Maybe just deprecate?

oleg-nenashev:
🐛 for removal of API class. Other code LGTM

jglick:
This is not an API plugin, and there should be no code using the deleted class. I do not intend to keep it around.

abayer:
:bee:

oleg-nenashev:
"Not an API plugin" really requires a kind of a global annotation in BCT :( We never know what plugin developers and Groovy script developers consider as API.

So I'm not convinced. Would be better to just deprecate the code

jglick:
If anyone is using this class—not in @jenkinsci according to search—then they should stop immediately. I do not care to clutter the code base with deprecated junk that was never used externally to begin with. We do take care to ensure legitimate APIs are kept compatible across plugin releases.

jglick:
> Groovy script developers

The only conceivable reason for such a script to be referring to this class would be to do an `instanceof` check, which will now fail, since the class is unused. Much better IMO to fail with a comprehensible error about an undefined symbol than to silently not work as intended.

oleg-nenashev:
You can just ignore my bug as a maintainer. I'm not 100% happy about the approach, bu I can live with it of course

PR #16, opened 2016-08-09
SettableFuture.setException will rethrow Error

jglick:
Noticed while investigating [JENKINS-26192](https://issues.jenkins-ci.org/browse/JENKINS-26192).

@reviewbybees


oleg-nenashev:
🐝

PR #17, opened 2016-08-22
[FIXED JENKINS-37477] - Docs and snippet generator work properly for triggers.

abayer:
[JENKINS-37477](https://issues.jenkins-ci.org/browse/JENKINS-37477)

- [x] Fix doc generation.
- [x] Fix snippet generator

Snippet generator still not working, so more work to do.

cc @reviewbybees esp @jglick @kwhetstone

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
@@ -135,7 +135,10 @@ public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
     public PipelineTriggersJobProperty reconfigure(@Nonnull StaplerRequest req, @CheckForNull JSONObject form) throws Descriptor.FormException {
         DescribableList<Trigger<?>,TriggerDescriptor> trigList = new DescribableList<>(Saveable.NOOP);
         try {
-            JSONObject triggerSection = form.getJSONObject("triggers");
+            JSONObject triggerSection = new JSONObject();
+            if (form != null) {
+                triggerSection = form.getJSONObject("triggers");
+            }
             trigList.rebuild(req, triggerSection, Trigger.for_(owner));
         } catch (IOException e) {
             // TODO: Not sure what form field would make sense here?
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
@@ -50,16 +50,19 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+@SuppressWarnings("rawtypes")
 public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
     private List<Trigger<?>> triggers = new ArrayList<>();
 
     @DataBoundConstructor
-    public PipelineTriggersJobProperty(List<Trigger<?>> triggers) {
+    public PipelineTriggersJobProperty(List<Trigger> triggers) {
         // Defensive handling of when we get called via {@code Descriptor.newInstance} with no form data.
         if (triggers == null) {
             this.triggers = new ArrayList<>();
         } else {
-            this.triggers = triggers;
+            for (Trigger t : triggers) {
+                this.triggers.add((Trigger<?>)t);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


abayer:
Note that I can't do a SNAPSHOT deploy any more thanks to changes to Artifactory configuration, so the PR I'm going to make on multibranch for the snippet generator fix eventually will fail. yay.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -50,16 +50,19 @@
...
-    public PipelineTriggersJobProperty(List<Trigger<?>> triggers) {
+    public PipelineTriggersJobProperty(List<Trigger> triggers) {
         // Defensive handling of when we get called via {@code Descriptor.newInstance} with no form data.
         if (triggers == null) {
             this.triggers = new ArrayList<>();
         } else {
-            this.triggers = triggers;
+            for (Trigger t : triggers) {
+                this.triggers.add((Trigger<?>)t);
+            }
```
jglick: Why?

jglick:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
abayer: Because `this.triggers = triggers;` doesn't work when `this.triggers` is `List<Trigger<?>>` and `triggers` is `List<Trigger>`. :hooray:

kwhetstone:
:bee:

oleg-nenashev:
:bee:

abayer:
Hmm, I don't think I saw any errors running locally. Retrying locally to make sure...

abayer:
I was wrong. There are errors. Fun!

abayer:
Yeah, so, dumb me, I looked at the wrong tab and was looking at test results before changing the config.jelly. Sigh! `reconfigure` is fixed now, so the tests should pass.

kwhetstone:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
jglick: You can use an unchecked cast, would reduce code changes.

src/main/resources/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty/config.jelly
```diff
@@ -31,9 +31,15 @@ THE SOFTWARE.
...
-    <f:descriptorList title="${%Build Triggers}"
+    <f:descriptorList field="triggers"
+                      title="${%Build Triggers}"
                       descriptors="${triggers}"
                       instances="${it.triggers}">
         <d:invokeBody />
     </f:descriptorList>
+    <f:block>
+        <div class="section-header">
+            ${%Other Properties}
```
jglick: Hmm, clumsy, inconsistent with freestyle projects, and fragile (you have no way of knowing for sure where this property will appear). I wonder if we could somehow make this property appear last in the list instead, using `ordinal`. Also a hack but at least would look better.

Alternately we could have a proper section for **Properties** (?) and then have **Build Triggers** be an indented block rather than a section…though this would again be inconsistent with traditional project types.

jglick:
re:bee:

src/main/resources/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty/config.jelly
abayer: Ordinal sounds like a decent approach to me. Yes, a hack, but... I'd like to eventually clean up the properties section entirely, but for now, a hack seems like the only option. Changing to `ordinal` now.

rsandell:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -132,26 +135,29 @@ public void removeTrigger(Trigger t) {
...
+        PipelineTriggersJobProperty thisProp = new PipelineTriggersJobProperty(new ArrayList<Trigger>(trigList.toList()));
         thisProp.setOwner(owner);
 
         thisProp.startTriggers(true);
 
         return thisProp;
     }
 
-    @Extension
+    @Extension(ordinal = -100)
```
jglick: Ugly but good enough for now I think.

jglick:
re:bee:

PR #19, opened 2016-09-06
[FIXED JENKINS-37872] Add WorkflowRun.getCulprits()

abayer:
[JENKINS-37872](https://issues.jenkins-ci.org/browse/JENKINS-37872)
- [x] Add a basic `getCulprits()` with test.
- [ ] Decide if we want to emulate `AbstractBuild.upstreamCulprits`.
- [ ] Determine if there's a way to limit culprits to just the repo the `Jenkinsfile` comes from if so desired?

FWIW: `upstreamCulprits` is built on top of `AbstractBuild.getDependencyChanges(AbstractBuild)`, which itself is dependent on `FingerprintAction` usage and `AbstractBuild.DependencyChange`, which is a whole complicated pile o' logic. So...good question as to whether that's functionality we want to emulate.

And as of right now, I honestly can't tell if there's a viable way to restrict the culprits to the SCM checkout that provided the `Jenkinsfile` - the `ChangeLogSet`s don't seem to preserve their origin SCM information...

cc @reviewbybees esp @jglick

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
@@ -39,23 +39,26 @@ import hudson.security.Permission;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
 import jenkins.model.CauseOfInterruption;
 import jenkins.model.InterruptedBuildAction;
 import jenkins.model.Jenkins;
+import jenkins.plugins.git.GitSampleRepoRule;
+import jenkins.plugins.git.GitStep;
 import org.apache.commons.io.FileUtils;
 import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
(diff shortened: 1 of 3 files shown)


abayer:
Ok, so this is still failing on CI but passes for me locally - apparently "git -c user.name='foo' -c user.email='foo@example.com' commit --all --message=something" isn't changing the committer from the default "nobody" on CI. Urk.

abayer:
And yes, I'd tried "git config user.name foo; git config user.email 'foo@example.com'" previously. That also worked locally on my Mac but not on CI.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -299,6 +301,8 @@ private void updateJenkinsfileWithCommitter(String committerName, String committ
     @Issue("JENKINS-37872")
     @Test
     public void culprits() throws Exception {
+        // TODO: Eventually change this to override GIT_COMMITTER_NAME etc. For now, skip if that's set.
+        Assume.assumeTrue("GIT_COMMITTER_NAME env var not set", System.getenv("GIT_COMMITTER_NAME") == null);
```
jglick: Good enough for now I guess.

jglick:
> whether that's functionality we want to emulate

I do not think we want to _emulate_ it, we want to _use_ it. I am a soft :-1: on duplicating code like this, especially when the copy does not really behave the same as the original. Core code needs to be refactored to offer APIs we can consume from a generalized `Job`.

> a viable way to restrict the culprits to the SCM checkout that provided the `Jenkinsfile`

Why would you want to do that? If the build broke, any participating commit might have been at fault.

abayer:
Closing this in favor of doing it right a la https://issues.jenkins-ci.org/browse/JENKINS-24141

PR #20, opened 2016-09-13
[JENKINS-28222] Branch labels were being incorrectly skipped after an end block node

jglick:
[JENKINS-28222](https://issues.jenkins-ci.org/browse/JENKINS-28222)

Pairs with https://github.com/jenkinsci/workflow-cps-plugin/pull/54.

@reviewbybees

PR #21, opened 2016-09-13
[JENKINS-27394] Collapsible sections in log

jglick:
[JENKINS-27394](https://issues.jenkins-ci.org/browse/JENKINS-27394)

Downstream of #20.

I initially thought to add visual bounding boxes, as the Collapsing Console Sections plugin does. But this would not work with `parallel` steps at all, and it seemed clumsy to show bounding boxes unless you were inside a branch.

A more radical alternative would be to _never_ show multiple branches interleaved at all, and have a special control at the start of a `parallel` branch allowing you to select which branch to display. This would usually be fine, but there are also some cases where you actually want to know what happened in which order; Pipeline captures this information and it would be a shame to throw it away. The likely replacement would be to bake in the equivalent of the `timestamper` wrapper so you could manually compare timestamps between branches in cases where this was important for debugging something.

Note a little log cleanup in https://github.com/jenkinsci/workflow-support-plugin/pull/10. The longer-term plan is to interleave all output into a single log file internally, with flow node (and perhaps timestamp) metadata captured at point of origin, and various visualizations like the annotated console log able to display the data however they prefer.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -885,7 +885,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         String prefix = getLogPrefix(node);
         String text = node.getDisplayFunctionName();
         if (node instanceof BlockStartNode) {
-            text += " (" + ShowHideNote.encodeTo(node.getId(), true, "show") + "/" + ShowHideNote.encodeTo(node.getId(), false, "hide") + ")";
+            text += " (" + ShowHideNote.create(node.getId()) + ")";
         }
         if (prefix != null) {
             wfLogger.log(String.format("[%s] %s", prefix, text));
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -61,6 +61,7 @@ import java.io.OutputStream;
 import java.io.PrintStream;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -94,8 +95,11 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.flow.GraphListener;
 import org.jenkinsci.plugins.workflow.flow.StashManager;
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
+import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
 import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.job.console.NestingNote;
+import org.jenkinsci.plugins.workflow.job.console.ShowHideNote;
 import org.jenkinsci.plugins.workflow.job.console.WorkflowConsoleLogger;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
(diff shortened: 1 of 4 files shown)


jglick:
Particularly useful in https://github.com/jenkinsci/parallel-test-executor-plugin/pull/20.

src/main/java/org/jenkinsci/plugins/workflow/job/console/ShowHideNote.java
```diff
@@ -0,0 +1,66 @@
...
+            // impossible, but don't make this a fatal problem
+            LOGGER.log(Level.WARNING, "Failed to serialize " + ShowHideNote.class, e);
+            return text;
+        }
+    }
+
+    private final String id;
+
+    private ShowHideNote(String id, int length) {
+        super("#", length);
```
jglick: Might as well use `"#" + id` here and add an `a` tag in `logNodeMessage`, allowing us to create direct links to the section—useful for `getLogURL` in https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/2, for example.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -452,6 +461,37 @@ private long writeRawLogTo(AnnotatedLargeText<?> text, long start, OutputStream
...
+                    @Override public @Nonnull List<String> load(FlowNode node) {
+                        if (node instanceof BlockEndNode) {
+                            return getNesting(((BlockEndNode) node).getStartNode());
+                        } else {
+                            List<FlowNode> parents = node.getParents();
+                            if (parents.isEmpty()) { // FlowStartNode
+                                return Collections.emptyList();
+                            }
+                            List<String> parent = getNesting(parents.get(0)); // multiple parents is only for BlockEndNode after parallel
+                            if (node instanceof BlockStartNode) {
```
jglick: :bug: produces incorrect results for

```groovy
stage('one') {}
stage('two') // should return singleton list; instead returns two stage nodes
{}
```

src/main/java/org/jenkinsci/plugins/workflow/job/console/ShowHideNote.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Shows or hides a block by nesting.
+ */
+public class ShowHideNote extends HyperlinkNote {
+
+    private static final Logger LOGGER = Logger.getLogger(ShowHideNote.class.getName());
+    private static final long serialVersionUID = 1L;
+
+    public static String create(String id) {
+        String text = "hide";
```
jglick: Perhaps hide all but the first branch of `parallel` by default?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -831,12 +871,26 @@ public int hashCode() {
...
+        }
+        try {
+            listener.annotate(new NestingNote(nesting));
+        } catch (IOException x) {
+            LOGGER.log(Level.WARNING, null, x);
+        }
         WorkflowConsoleLogger wfLogger = new WorkflowConsoleLogger(listener);
         String prefix = getLogPrefix(node);
+        String text = node.getDisplayFunctionName();
+        if (node instanceof BlockStartNode) {
```
jglick: Would be useful to be able to collapse `AtomNode` output as well: for example, long `sh` steps, or `build` if [JENKINS-26124](https://issues.jenkins-ci.org/browse/JENKINS-26124) is implemented.

jglick:
Putting this on hold until [JENKINS-38381](https://issues.jenkins-ci.org/browse/JENKINS-38381) is done.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: @jglick Consider using https://github.com/jenkinsci/workflow-api-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/SimpleChunkVisitor.java  to collect ForkScanner results - you can easily roll up the result with a simple ArrayDequeue of blocks.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NestingNote.java
```diff
@@ -0,0 +1,60 @@
...
+import hudson.MarkupText;
+import hudson.console.ConsoleAnnotator;
+import hudson.console.ConsoleNote;
+import hudson.model.Run;
+import java.util.List;
+
+/**
+ * Encodes the block-scoped nesting of a step.
+ */
+public class NestingNote extends ConsoleNote<Run<?,?>> {
```
jglick: :bug: needs to be `Object` to avoid breaking per-step `LogAction` display

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: For this case I want incremental caching, which AFAICT is not yet supported.

jglick:
Closing in favor of #27.

PR #23, opened 2016-09-13
[JENKINS-33821] Pipeline jobs can't be disabled (alternate)

csiden:
This goes together with the Jenkins core change in https://github.com/jenkinsci/jenkins/pull/2544. While the core change is related, it is not required to use the disable/enable functionality in this diff. I have tested this change with Jenkins core 1.642.3 (and with the pull request above) and disable/enable seems to work, including cancelling queued jobs, updating the job BallColor, blocking SCM polling, and blocking cron triggers. The core functionality of "don't do builds if a job is disabled" seems to work as long as WorkflowJob.isBuildable() returns 'false' when the job is disabled, which does not depend on the Jenkins core change.

However, because Jenkins core and plugins use AbstractProject.isDisabled() rather than Job<?,?>.isDisabled() (introduced in https://github.com/jenkinsci/jenkins/pull/2544) there may be some odd interactions until other plugins are updated, since they won't be able to tell when pipeline jobs are disabled.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -661,11 +661,16 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements B
     }
 
     private void makeDisabled(boolean b) throws IOException {
-        if (disabled == b)
+        if (disabled == b) {
             return;
+	}
         this.disabled = b;
-        if (b)
-            Jenkins.getInstance().getQueue().cancel(this);
+        if (b) {
+	    Jenkins instance = Jenkins.getInstance();
+	    if (instance != null) {
+		instance.getQueue().cancel(this);
+	    }
+	}
         save();
     }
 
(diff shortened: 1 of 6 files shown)


csiden:
@jglick Is just opening pull requests here the correct way to submit changes for this plugin? Thanks, Chris.

jglick:
Yes though for complex changes it is best to check with me in IRC or whatever first. I have to think about whether the approach taken here is really what we want, or if a `JobProperty` is needed. Right now I cannot spare the time to review this properly.

csiden:
Okay, this didn't take long to put together so I didn't bother asking first, but let me know when you have time to take a look, I'm happy to redo it in whatever way is most correct.

jglick:
Looks reasonable; I need to think a bit about how this will interact with multibranch. As written you would be able to manually enable/disable branch projects, but not do so via `Jenkinsfile`. Which _may_ be for the best.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -632,7 +674,52 @@ public void replaceAction(Action a) {
...
+    }
+
+    /*
+     * @Overrides method added to Job<?, ?> in Jenkins core version TODO
+     */
+    public void enable() throws IOException {
+        makeDisabled(false);
+    }
+
+    @CLIMethod(name="disable-job")
```
jglick: Have to check what actually happens here. From a quick reading of `CLIRegisterer`, looks like either this command will be broken for `WorkflowJob` or it will be broken for `AbstractProject`. Certainly deserves to be tested at least.

jglick:
I plan to reimplement this.

jglick:
Superseded.

PR #25, opened 2016-09-19
WorkflowJob.setDefinition should save the project

jglick:
Simplifies tests using `RestartableJenkinsRule` which rebuild the project after a restart.

Would be more natural to throw up the `IOException`, but that would be source-incompatible for [this call](https://github.com/jenkinsci/workflow-multibranch-plugin/blob/4fa8a41c30a5e950d15df98df8126bb48ac8485b/src/main/java/org/jenkinsci/plugins/workflow/multibranch/AbstractWorkflowBranchProjectFactory.java#L63).

@reviewbybees


svanoort:
🐝 and if I could make it a swarm I would.

oleg-nenashev:
🐝

PR #26, opened 2016-09-19
Introduce copyLogsExecutorService to avoid blocking Timer

jglick:
@cyrille-leclerc found a case where all ten `Timer` threads were blocked in `WorkflowRun.copyLogs`, either inside `LogActionImpl.getLogText` or `writeRawLogTo`. The cause is unknown—perhaps an NFS outage. At any rate, it is unwise to block `Timer` indefinitely, because that prevents other things in the system from progressing, notably `DurableTaskStep.Execution.check`. This patch ensures that log copying is segregated into its own thread pool, which could bog down but not break anything else.

(`check` imposes a timeout on itself. I thought of doing that here as well, but the I/O calls seem uninterruptible: `FileInputStream.close0` etc.)

@reviewbybees


oleg-nenashev:
🐝

andresrc:
:bee:

PR #27, opened 2016-09-23
[JEP-210] Log handling rewrite

jglick:
Downstream of jenkinsci/workflow-api-plugin#17 and https://github.com/jenkinsci/workflow-support-plugin/pull/15. Cf. #21.

[JEP-210](https://github.com/jenkinsci/jep/blob/master/jep/210/README.adoc)

- [X] basics
- [x] avoid overhead of serialized `ConsoleNote` on every line
- [x] pluggable log location
- [x] tests of annotation, single-step log display
- [X] basic show/hide UI for steps (adapted from #21)
- [x] show/hide entire sections
- [ ] ~optimize `PipelineLargeText.<init>`~
- [ ] ~optimize `WorkflowRun.getLogInputStream`~

Nice-to-have things I am deferring:

- [ ] node anchors, e.g. for https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/2
- [ ] hide all but first `parallel` branch by default
- [ ] gesture to hide all Pipeline steps


pom.xml
```diff
@@ -52,40 +52,51 @@
...
     <pluginRepositories>
         <pluginRepository>
             <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
+            <url>https://repo.jenkins-ci.org/public/</url>
         </pluginRepository>
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
+        <hpi-plugin.version>1.120</hpi-plugin.version>
```
oleg-nenashev: Ideally should be updated in the Parent POM

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -200,8 +181,10 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
         }
         try {
             onStartBuilding();
+            // TODO introduce API to wrap out log location
```
oleg-nenashev: Implementation from the prototype: https://github.com/oleg-nenashev/workflow-job-plugin/tree/jw-hackaton-external-logging

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -825,6 +646,10 @@ private String key() {
                 return TaskListener.NULL;
             }
         }
+        @Override public InputStream getLog() throws IOException {
```
oleg-nenashev: Maybe Stream is not a perfect solution for Logging. I would rather go after an event-based model similar to JUL, but it will require more changes (it's in my design btw)

src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowRunConsoleNote.java
```diff
@@ -24,23 +24,22 @@
...
  * Console note for Workflow metadata specific messages.
- * See {@link WorkflowConsoleLogger} for more information.
  */
 public class WorkflowRunConsoleNote extends ConsoleNote<Run<?, ?>> {
 
     /**
      * Prefix used in metadata lines.
      */
-    public static final String CONSOLE_NOTE_PREFIX = "[Pipeline] ";
+    private static final String CONSOLE_NOTE_PREFIX = "[Pipeline] ";
```
oleg-nenashev: 🐜 binary compat change without a real need

pom.xml
jglick: Please review https://github.com/jenkinsci/plugin-pom/pull/35. :-)

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Saw that, but it is not going to work.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -603,6 +417,13 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
                 completed.set(true);
             }
         }
+        Executor executor = getExecutor();
+        if (executor != null) {
+            AsynchronousExecution asynchronousExecution = executor.getAsynchronousExecution();
+            if (asynchronousExecution != null) {
+                asynchronousExecution.completed(null);
```
jglick: Uses to be part of `copyLogsTask` (line 278); that is now gone, so we need to somehow release the flyweight executor slot. This seems like the appropriate place to do it.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Tons of Jenkins APIs expect a `TaskListener` or `OutputStream` for output, and expect an `AnnotatedLargeText` or `InputStream` for input, so streams are mandated at some level I think. Of course a logging implementation might internally work with lines.

src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowRunConsoleNote.java
jglick: Just minimizing access of things which were never really supposed to be `public`.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
```diff
@@ -0,0 +1,125 @@
...
+        try {
+            listener.annotate(new NewNodeConsoleNote(node));
+        } catch (IOException x) {
+            // never mind
+        }
+        listener.getLogger().println(CONSOLE_NOTE_PREFIX + node.getDisplayFunctionName());
+    }
+
+    private final @Nonnull String id;
+    private final @Nonnull String[] parents;
```
jglick: Perhaps just replace with `@CheckForNull String enclosing` (mutually exclusive with `start`) so that `script.js` does not need to reimplement `FlowScanningUtils.fetchEnclosingBlocks`.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
```diff
@@ -0,0 +1,131 @@
...
+        FlowExecution execution = context.getExecution();
+        if (execution != null) {
+            try {
+                FlowNode node = execution.getNode(id);
+                if (node != null) {
+                    LabelAction a = node.getAction(LabelAction.class);
+                    if (a != null) {
+                        String displayName = a.getDisplayName();
+                        assert displayName != null;
+                        startTag.append("\" label=\"").append(Util.escape(displayName)); // TODO is there some better way to escape for attribute values?
```
svanoort: Noting here for future review/integration that this appears to be how we propose to deal with rendering of branch labels -- however it requires rich rendering of annotation content by the UI to be useful (i.e. not visible in plaintext logs). 

Is that the right decision?   Not sure, the mechanism seems more straightforward in some ways.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
```diff
@@ -0,0 +1,131 @@
...
+    private final @CheckForNull String enclosing;
+    private final @CheckForNull String start;
+
+    private NewNodeConsoleNote(FlowNode node) {
+        id = node.getId();
+        if (node instanceof BlockEndNode) {
+            enclosing = null;
+            start = ((BlockEndNode) node).getStartNode().getId();
+        } else {
+            Filterator<FlowNode> it = FlowScanningUtils.fetchEnclosingBlocks(node).filter(Predicates.not(Predicates.equalTo(node)));
```
svanoort: Future review note: replace with the FlowNode APIs to get enclosing blocks now (much faster lookup).

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
jglick: Right, this keeps the HTML leaner and does more of the fanciness on the client side.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
jglick: Yeah I could probably make that change easily, good catch.

jglick:
> On loading in-progress builds

Specifically applies just to _steps_ running across the update. Those completed prior to the update should already have been copied to the main log. (And will be using `LogActionImpl`, so subsequent lookups of the step output via `LogAction` should work as before.) Those started after the update should stream to the main log. Since this is a one-off task applying only to a limited number of steps and thus not a performance bottleneck, we can probably have a simplified implementation rather than the current `logsToCopy` mess. This might suffice (but needs test coverage):

```diff
diff --git a/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java b/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
index c1a0611..93eef32 100644
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.workflow.support.actions;
 
 import com.google.common.base.Charsets;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.CloseProofOutputStream;
 import hudson.console.AnnotatedLargeText;
 import hudson.console.ConsoleLogFilter;
 import hudson.model.AbstractBuild;
@@ -43,6 +44,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import org.apache.commons.io.output.TeeOutputStream;
 import org.apache.commons.jelly.XMLOutput;
 import org.jenkinsci.plugins.workflow.actions.FlowNodeAction;
 import org.jenkinsci.plugins.workflow.actions.LogAction;
@@ -78,6 +80,7 @@ public class LogActionImpl extends LogAction implements FlowNodeAction, Persiste
             node.addAction(la);
         }
         OutputStream os = new FileOutputStream(la.getLogFile(), true);
+        os = new TeeOutputStream(os, new CloseProofOutputStream(node.getExecution().getOwner().getListener().getLogger()));
         if (filter != null) {
             os = filter.decorateLogger((AbstractBuild) null, os);
         }
```

jglick:
> This might suffice

FTR this was already done in https://github.com/jenkinsci/workflow-support-plugin/pull/15/commits/468f91dca27d0f2c2b922489cd289802980b7906.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
```diff
@@ -31,7 +31,17 @@
             <l:task icon="images/24x24/up.png" href="${rootURL}/${it.parent.url}" title="${%Back to Project}" contextMenu="false"/>
             <l:task icon="images/24x24/search.png" href="${buildUrl.baseUrl}/" title="${%Status}" contextMenu="false"/>
             <l:task icon="images/24x24/notepad.png" href="${buildUrl.baseUrl}/changes" title="${%Changes}"/>
-            <p:console-link/>
+            <j:choose> <!-- TODO <p:console-link/> may not currently be used as it calls getLogFile -->
+                <j:when test="${it.logText.length() > 200000}">
```
jglick: Perhaps `LogStorage` needs a new API like

```java
default int approximateLogLines(@Nonnull FlowExecutionOwner.Executable build) {
    return overallLog(build, false).length() / 80;
}
```

which could be overridden by external implementations to make an API call.

Of course there is the broader goal of [JENKINS-17406](https://issues.jenkins-ci.org/browse/JENKINS-17406) to use paginated display.

jglick: Specifically:

* `console-link.jelly` uses `it.logFile.length()`, which for `WorkflowRun` now prints a warning—very bad. Thus we specifically avoid this control until core is fixed. Fortunately it is not much to copy.
* `console.jelly` uses `it.logText.length()`, which is fine for the built-in `FileLogStorage`: boils down to a single `File.length()` call after verifying that the first few bytes of the file are not a gzip magic header.
* For `pipeline-log-fluentd-cloudwatch` and the like, this is very inefficient, and we would prefer an API for an approximation. (On reflection this need not be in `LogStorage`; could be in `LargeText` or `AnnotatedLargeText`.)

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
jglick: AFAICT `FlowScanningUtils.fetchEnclosingBlocks` just delegates to `FlowNode.iterateEnclosingBlocks`, so the performance will be the same, just avoids a deprecation warning.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1277,6 +1110,15 @@ public void save() throws IOException {
         synchronized (this) {
             PipelineIOUtils.writeByXStream(this, loc, XSTREAM2, isAtomic);
             SaveableListener.fireOnChange(this, file);
+            if (Boolean.TRUE.equals(completed)) {
+                Executor executor = getExecutor();
+                if (executor != null) {
+                    AsynchronousExecution asynchronousExecution = executor.getAsynchronousExecution();
+                    if (asynchronousExecution != null) {
+                        asynchronousExecution.completed(null);
+                    }
```
dwnusbaum: Maybe log a warning if `asynchronousExecution`  is null? I don't think it should ever happen, but we've seen some bugs where OneOffExecutors stick around as zombies even when their associated Pipelines appear to have completed (e.g., [JENKINS-50199](https://issues.jenkins-ci.org/browse/JENKINS-50199), [JENKINS-45571](https://issues.jenkins-ci.org/browse/JENKINS-45571)), so it could end up being useful one day.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
```diff
@@ -0,0 +1,133 @@
...
+     */
+    private static final String CONSOLE_NOTE_PREFIX = "[Pipeline] ";
+
+    public static void print(FlowNode node, TaskListener listener) {
+        PrintStream logger = listener.getLogger();
+        synchronized (logger) {
+            try {
+                listener.annotate(new NewNodeConsoleNote(node));
+            } catch (IOException x) {
+                // never mind
```
dwnusbaum: Is it worth logging the exception at Level.FINE (or higher?) in case we are debugging something and suspect issues here?

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
```diff
@@ -0,0 +1,133 @@
...
+        } else {
+            Iterator<BlockStartNode> it = node.iterateEnclosingBlocks().iterator();
+            enclosing = it.hasNext() ? it.next().getId() : null;
+            start = node instanceof BlockStartNode ? node.getId() : null;
+        }
+    }
+
+    @Override
+    public ConsoleAnnotator<?> annotate(WorkflowRun context, MarkupText text, int charPos) {
+        StringBuilder startTag = new StringBuilder("<span class=\"pipeline-new-node\" nodeId=\"").append(id);
```
dwnusbaum: nit: Why not close each attribute immediately (e.g. `new StringBuilder("<span class=\"pipeline-new-node\" nodeId=\"").append(id).append("\"")`) to make it more obvious to readers that they are closed correctly?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Probably unrelated.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
jglick: Currently it is impossible AFAIK. `PrintStream` will not throw exceptions, and the note should never fail to serialize to the `ByteArrayOutputStream`.

(Most `ConsoleNote`s have a similar utility method.)

Anyway, I can log it just in case.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
jglick: Merely because that involves more concatenations total.

src/test/java/org/jenkinsci/plugins/workflow/job/CpsPersistenceTest.java
```diff
@@ -102,6 +102,7 @@ static void assertCompletedCleanly(WorkflowRun run) throws Exception {
             Assert.assertTrue(cpsExec.getCurrentHeads().get(0) instanceof FlowEndNode);
             Stack<BlockStartNode> starts = getCpsBlockStartNodes(cpsExec);
             Assert.assertTrue(starts == null || starts.isEmpty());
+            Thread.sleep(1000); // TODO seems to be flaky
```
jglick: From #104.

jglick:
Also added an automated upgrade test which exercises the compatibility code in `LogActionImpl` + `LogStorageAction`. The first half of the build, including part of the `sleep` step, runs in current `master` (plus its test dependencies); the second, including the rest of the `sleep` step, in this branch. The new console UI takes effect only on steps started after the upgrade (or, more precisely, which start producing output for the first time after the upgrade):

upgrade

dwnusbaum:
I was testing out this PR and the upstream changes in workflow-api and workflow-support interactively, and I noticed that when I view the logs of pipeline builds I ran before upgrading those plugins, I see Base64-encoded console notes literally in the output. Is that expected?

jglick:
> when I view the logs of pipeline builds I ran before upgrading all 3 plugins, I see Base64-encoded console notes literally in the overall log. Is that expected?

No, I do not recall seeing anything like that. Reproducible?

jglick:
OK, I just saw this for the first time, I will track down what is going on and fix it.

jglick:
Will solve upstream.

jglick:
> I see multiple hide/show links next to steps for running builds in the log

If true that is a bug. I had not seen such an issue.

jglick:
> I had not seen such an issue.

Nor can I reproduce such an issue when using the versions stated in this PR (`mvn hpi:run`), either in a build started in this session and paused; or after resuming the build after a Jenkins restart; or after the build completes and the page is reloaded.

dwnusbaum:
> Nor can I reproduce such an issue ... after the build completes and the page is reloaded

Note that the issue was only visible while the build was running, but I will try again to reproduce it (maybe I had mismatched dependencies).

dwnusbaum:
@jglick I can still reproduce the issue after cleaning out my work directory and running the plugin using `mvn hpi:run`. It only occurs while the build is running - once the build completes the second link is gone. Also, it looks like the issue is not visible immediately for the first build after Jenkins starts/restarts, and you have to refresh the page for the second link to show up. My pipeline is just `sleep 100`.

screen shot 2018-09-24 at 12 03 24

jglick:
> you have to refresh the page for the second link to show up

Ah, this was the critical bit to reproduce. I will take a look.

jglick:
Somehow [this code](https://github.com/jenkinsci/jenkins/blob/cb6b5ea7f3baae3313e4ca11ee5cda84bb1d9eaa/core/src/main/resources/lib/hudson/progressiveText.jelly#L75) causes behaviors to be applied twice under some conditions (the second call coming [here](https://github.com/jenkinsci/jenkins/blob/cb6b5ea7f3baae3313e4ca11ee5cda84bb1d9eaa/war/src/main/webapp/scripts/behavior.js#L125)). It is unclear to me whether this is a bug in core, or whether behaviors are simply expected to be idempotent. At any rate, the fix looks easy enough.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -813,10 +596,16 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
...
+                        ((AutoCloseable) listener).close();
+                    } catch (Exception x) {
+                        LOGGER.log(Level.WARNING, "could not close build log for " + this, x);
+                    }
+                }
+                listener = null;
             }
-            logsToCopy = null;
-            saveWithoutFailing();
+            saveWithoutFailing(); // TODO useless if we are inside a BulkChange
```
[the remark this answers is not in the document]
dwnusbaum: Yeah I don't consider this a blocker, just something to keep in mind if we see issues with Pipelines not completing after this change. 

My initial thought for ensuring completion was to add a new `WorkflowRun#save` overload that takes a new `force` parameter, and if that is true then when `#save` returns early because of a bulk change, it sets a new transient `WorkflowRun.needsToSave` field to `true` (the field would be set to false at the end of `#save`) and schedules a background task to call `#save` again in a few seconds if that field is not `false`. Either way I think we can leave this for now and come back to it if we see any problems.

src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowConsoleLogger.java
```diff
@@ -1,75 +0,0 @@
...
-
-/**
- * Console logger used to write workflow related metadata in console text.
- *
- * It wraps the regular {@link BuildListener} to add a filter that annotates any text line sent to the console through
- * this logger. It also adds a prefix to the line ([Pipeline]).
- *
- * Annotated lines will be rendered in a lighter color so they do not interefere with the important part of the log.
- */
-public class WorkflowConsoleLogger {
```
svanoort: 🐜 Not binary compatible change.  But I can't find a reference to it being used anywhere, so... 🤷‍♂️

src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowMetadataConsoleFilter.java
```diff
@@ -1,55 +0,0 @@
...
-
-import java.io.IOException;
-import java.io.OutputStream;
-
-import hudson.console.LineTransformationOutputStream;
-
-/**
- * It transforms workflow metadata log messages through {@link WorkflowRunConsoleNote}.
- */
-public class WorkflowMetadataConsoleFilter extends LineTransformationOutputStream {
```
svanoort: Also technically binary incompatible, but meh

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1226,19 +1005,73 @@ public int hashCode() {
...
+
+    @SuppressWarnings("rawtypes")
+    @Override public AnnotatedLargeText getLogText() {
+        return LogStorage.of(asFlowExecutionOwner()).overallLog(this, !isLogUpdated());
+    }
+
+    // TODO log-related overrides pending JEP-207:
+
+    @Override public InputStream getLogInputStream() throws IOException {
+        // Inefficient but probably rarely used anyway.
```
svanoort: And comments like this make me realize that the log-handling rewrite that was supposed to deliver some great performance improvements may fall well short of that mark.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1226,19 +1005,73 @@ public int hashCode() {
...
+    @SuppressWarnings("deprecation")
+    @Override public String getLog() throws IOException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        getLogText().writeRawLogTo(0, baos);
+        return baos.toString("UTF-8");
+    }
+
+    @Override public List<String> getLog(int maxLines) throws IOException {
+        int lineCount = 0;
+        List<String> logLines = new LinkedList<>();
```
svanoort: 🐜 [Why you should never use a LinkedList](https://kjellkod.wordpress.com/2012/02/25/why-you-should-never-ever-ever-use-linked-list-in-your-code-again/).  TL;DR: Locality-of-reference matters, allocate a darned `ArrayList`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1226,19 +1005,73 @@ public int hashCode() {
...
+        if (maxLines == 0) {
+            return logLines;
+        }
+        try (BufferedReader reader = new BufferedReader(getLogReader())) {
+            for (String line = reader.readLine(); line != null; line = reader.readLine()) {
+                logLines.add(line);
+                ++lineCount;
+                if (lineCount > maxLines) {
+                    logLines.remove(0);
```
svanoort: 🐛 And `break`!  Don't keep adding-and-removing. 

Better yet, use `line != null && lineCount <= maxLines` as part of the `for` loop conditions

svanoort: Actually... better expressed as:

```
while((line = reader.readLine()) != null && ++lineCount <= maxLines) {
    logLines.add(line)
}
```

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1226,19 +1005,73 @@ public int hashCode() {
...
+                }
+            }
+        }
+        if (lineCount > maxLines) {
+            logLines.set(0, "[...truncated " + (lineCount - (maxLines - 1)) + " lines...]");
```
svanoort: 🐛 You just disappeared the first line of log output.  I think you mean to *insert* there.  And the method should note that it actually returns maxLines+1 lines due to the prefix.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1291,6 +1124,15 @@ public void save() throws IOException {
         synchronized (this) {
             PipelineIOUtils.writeByXStream(this, loc, XSTREAM2, isAtomic);
             SaveableListener.fireOnChange(this, file);
+            if (Boolean.TRUE.equals(completed)) {
```
svanoort: @jglick No! 🐛 🐛 🐛 and a whole nest of 🐛s.   This is not the right place to be terminating the AsynchronousExecution - if an IOException never occurs, the build never finishes. 

You're welcome to do an extra on-save check _somewhere logical_ -- because we have had hard-to-trace issues with the build not getting killed before, but this makes undead Pipelines more likely.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -540,173 +492,6 @@ public boolean hasAllowKill() {
         return isBuilding() && allowKill;
     }
 
-    @GuardedBy("logCopyGuard")
-    private void copyLogs() {
```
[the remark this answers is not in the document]
jglick: Yes we do, and test coverage for it.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: My interest is in writes, not reads.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: This is just copied from core.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Ditto, all copied from core, slated to be removed in JEP-207 as the comment above says.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Ditto, copied from core.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: There is explanation elsewhere of why we _have_ to do the termination here. I can think about whether it is possible to do so more forcefully, such as in a `finally` block.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
jglick: `stepRunningAcrossUpgrade` FTR

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
jglick: > a new transient `WorkflowRun.needsToSave` field to `true` (the field would be set to false at the end of `#save`) and schedules a background task to call `#save` again in a few seconds if that field is not `false`

Should not be necessary if `BulkChange` is being used correctly with a `finally` block: sooner or later, in this thread, you are `save`ing. I have not followed the issues in https://github.com/jenkinsci/workflow-cps-plugin/pull/234 closely and perhaps there is an outstanding bug in `workflow-cps` in this regard.

At any rate, the behavior here is: whenever we do first successfully save `build.xml` after the end of the build (i.e., we are for the first time writing `<completed>true</completed>`), `AsynchronousExecution.completed` will be called. Previously this was done at a more random time (up to 3s later IIRC) in `copyLogsTask`.

Note that

> Pipelines not completing

is not likely what you would observe: the `AsynchronousExecution` corresponds only to the display of an unnumbered (flyweight) “executor slot” under **master** in the **Build Executor Status** widget, which is usually hidden (only rendered while the CPS VM thread is active). It has no effect on whether the build is considered completed for any other purpose, such as the **Build History** widget etc.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Note that the core implementation of _this_ method has since been rewritten. As I pointed out in https://github.com/jenkinsci/jenkins/pull/2607#issuecomment-425475683 that is irrelevant, since this method is not used in production workflows. Anyway the rewrite is impossible to apply here since it relies on `RandomAccessFile`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: > if an IOException never occurs

I suppose you mean if an `IOException` _does_ occur.

> the build never finishes

As noted in a separate thread, even if no subsequent call to `save` is ever made, it is misleading to say that the build is “not finished”. Rather, there would be a stale `AsynchronousExecution`, which likely has little or no visible impact.

src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowConsoleLogger.java
jglick: It was only ever `public` for technical reasons; not an API.

src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowMetadataConsoleFilter.java
jglick: Ditto.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
jglick: Needs to be revisited when resuming work on external log sinks. I am no longer working on that.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: (see https://github.com/jenkinsci/workflow-job-plugin/pull/27#discussion_r221294107 for fuller discussion)

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
dwnusbaum: Yes, I should clarify that I mean flyweight executors hanging around after the build completes rather than the build completion itself. The issue I am thinking of ([JENKINS-45571](https://issues.jenkins-ci.org/browse/JENKINS-45571)) was noticed by people looking at API output and seeing OneOffExecutors for long-completed jobs (which seems somewhat harmless though misleading), and by people using the Throttle Concurrent Builds Plugin where the flyweights count towards concurrent builds (not so harmless).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Copy and pasted code?  Still... 

I'm going to keep pointing out that LinkedList is a bad idea until people stop (ab)using it.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: If it's wrong in core, it can be fixed in core too.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: @jglick A bug that exists in TWO places is even worse than a bug existing in one, not better...

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: A stale AsynchronousExecution translates to essentially a "zombie" job that proves highly irritating to users -- it seems like it never completes even if the build is marked as done.  Also it holds onto the OneShotExecutor indefinitely.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Or is there a different argument you're trying to make here?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: As noted elsewhere, the newer implementation in core is different, and unusable here.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: This code is _temporarily_ copied from core pending JEP-207 or some other patch to core to delegate to `getLogText`. As soon as core does proper delegation, this override can be removed, so there is no reason to waste time fixing (supposed) buglets in code which has existed for years and is not really part of this PR. Anyway this method is never used in production code, only test output.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Again the `AsynchronousExecution` normally has no visible effect on anything—it only appears while we are actively running Groovy code, which of course we are not if the build is `completed`.

And again if some problem in _existing_ code (whether in `workflow-job` or `workflow-cps`) results in a condition whereby `<completed>true</completed>` never gets written to `build.xml`, the `AsynchronousExecution` is the least of your concerns, since after the next restart `onLoad` will be confused. Possibly #93 will defend against that case—figure out that `completed` is wrong and save the correction—but at that point any old `Executor`s are no longer an issue anyway (and there should be no new `AfterRestartTask`).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
jglick: > flyweights count towards concurrent builds

If true, that would be a bug in that plugin I think.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
dwnusbaum: The Javadoc for `Run#maxLines` says:

> The maximum number of log lines to return.  If the log is bigger than this, only the most recent lines are returned.

Given that documentation, it doesn't seem like returning `maxLines-1` of log content and one line explaining that things were truncated is a problem. Besides, in `Run#getLog` there is [a comment](https://github.com/jenkinsci/jenkins/blob/c4acc81405068fd969272fdb56c19da161a7015b/core/src/main/java/hudson/model/Run.java#L2079) explaining that `List#set` is used intentionally to avoid returning more lines:

> // If the log has been truncated, include that information.
> // Use set (replaces the first element) rather than add so that
> // the list doesn't grow beyond the specified maximum number of lines.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
jglick: > The issue I am thinking of ([JENKINS-45571](https://issues.jenkins-ci.org/browse/JENKINS-45571))

It does not sound like there is a definite way to reproduce that (?) but I see this JIRA comment:

> the build completes (`WorkflowRun#finish` is called), but the `copyLogsTask` never gets invoked or fails, since that is what actually removes the `FlyWeightTask` and kills the `OneOffExecutor`

This PR of course rewrites that logic, again with an explanation in 18d78f305a4526af9cdf3a7b68eb9caf97c7cfbc. We are no longer relying on the `copyLogs` system, which IIRC could sometimes get blocked on thread pool starvation, though perhaps this was bypassed by `GraphL` calling `copyLogs` directly. At any rate, now that call is guaranteed after `WorkflowRun.finish` sets `completed = true` so long as the `build.xml` gets saved _sometime_.

As to when that happens, since `finish` is not called from inside a `GraphListener.Synchronous`, the change in https://github.com/jenkinsci/workflow-cps-plugin/pull/234 should have no effect here AFAICT: with or without that, we are calling `BulkChange.commit` in a `finally`-block, which should result in `save` proceeding. So the only way you have a `FutureImpl` declining to consider the build complete (until the next restart) is if `build.xml` in fact still claims that the build is _not_ complete.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
dwnusbaum: I think the after-restart case should have been fixed by #104.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
dwnusbaum: Yes, there is no way to reproduce that issue that we know of, and that comment was our best hypothesis of the root cause. Like I said before, I am fine with the change, as it seems much easier to diagnose any issues with this implementation than it would be with `copyLogs`, and we can always revisit it if needed.

(Re: throttle-concurrent-builds: I think that behavior is intentional based on [a comment and the history](https://github.com/jenkinsci/throttle-concurrent-builds-plugin/blame/521b885be274b9a4fba13780dfaeb94891539051/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java#L486), but I have not looked into it deeply.)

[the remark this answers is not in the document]
svanoort: The dangling OneOffExecutors have practical and irritating implications for tests, memory use, and garbage-collectibility of build records too.

The AsynchronousExecution should be terminated at the time we mark the completed flag. 

Integrating this aspect with the already complex (and unfortunately still somewhat failure-prone) persistence model is an exceedingly poor idea.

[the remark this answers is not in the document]
jglick: > The AsynchronousExecution should be terminated at the time we mark the completed flag.

That was what I did originally. It did not work well for reasons explained in 18d78f305a4526af9cdf3a7b68eb9caf97c7cfbc. You need to wait until the state is actually written.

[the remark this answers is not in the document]
svanoort: Looking at that commit in depth, this choice makes *more* sense -- you basically wanted to guarantee it actually was saved before killing the AsynchronousExecution due to restarts.  But the catch there is we can't *guarantee* saves will work (disk failure, etc etc), and we do want to guarantee the AsynchronousExecution is completed when the build is. 

The on-restart logic is handled by #104 and that should do the bulk of needed cleanup for us if the persist-through-to-disk fails.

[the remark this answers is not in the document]
svanoort: We handle cases where the state is not written now though (#104) on restart. 

I'd say that if you really want to you could do something like this in `finish`:

```
try{
   if (BulkChange.contains(this)) {
         BulkChange.commit(this);
   }
   saveWithoutFailing()
} finally {
    `AsynchronousExecution.complete()`
}
```

(Note that I am writing from rough memory here, so I'm sure there are subtle typos in method signatures above, but you all get the gist).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Not much point of a log that nobody reads though, is there?  Anyway, we can see if it comes up as an issue.

Also, please do not mark this one "resolved" unless the code has been changed due to feedback -- it'll make it easier to address this if it come sup as a problem in the future.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: The actual explanation for the somewhat crazy logic here was done by Devin: https://github.com/jenkinsci/workflow-job-plugin/pull/27/files#r221330578 -- makes sense in light of that.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: This particular method is rarely called AFAIK.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: This method is not used in production. The performance does not matter.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
basil: To give the perspective of a Jenkins user, I was affected by the abovementioned issue of flyweight executors hanging around after completed builds. I use the Throttle Concurrent Builds Plugin, and the flyweight executors counted towards the concurrent builds throttle (preventing new builds from running). This took some time to debug and delayed the release of one of my company’s products.

If upgrading to a build with JEP-210 means there is a risk of this type of problem reoccurring, I would be very hesitant to upgrade. As a Jenkins administrator, I am primarily focused on delivering a stable CI solution to the other engineers at my company (which is why I always use LTS releases of Jenkins core). If JEP-210 carries some degree of risk, perhaps this could be explicitly called out in the plugin release notes or update sites (with a stable version provided as an alternative).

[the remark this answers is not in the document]
svanoort: @basil Definitely appreciate the comment here to give some context on why this is important -- I swear we worked together to pin down some of the prior issues, and that's part of why this particular issue was in mind.  In this case, I'm working on a patch to amend this branch which provides stronger protection against the issue you mention.

That said, this release will carry extra risks simply because it's such a major change, despite our best efforts to harden it against bugs.  I'm not sure if we've finalized the release plan yet, since it might go out as a beta first (I hope this will be the case).

@jglick I'd encourage these log-handling-rewrite releases carry an extra warning in the changelog -- bolded and underlined preferably -- because we all know that despite our best efforts there is a higher than normal risk of regressions since this change touches a lot of sensitive pieces.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public static void print(FlowNode node, TaskListener listener) {
+        PrintStream logger = listener.getLogger();
+        synchronized (logger) {
+            try {
+                listener.annotate(new NewNodeConsoleNote(node));
+            } catch (IOException x) {
+                LOGGER.log(Level.WARNING, null, x);
+            }
+            logger.println(CONSOLE_NOTE_PREFIX + node.getDisplayFunctionName()); // note that StepAtomNode will never have a LabelAction at this point
```
jglick: This fails `StepNodeTest`. Investigating.

jglick: Actually the bug is in `workflow-cps`’ `DSL.invokeStep` but I am looking for the right way to solve it.

PR #28, opened 2016-10-09
[FIXED JENKINS-38454] Make sure we add the triggers before starting.

KostyaSha:
https://issues.jenkins-ci.org/browse/JENKINS-38454
abayer: 
```
This was...complex. But it works. We make sure the triggers property
has been added before starting, in case the start method for a trigger
class expects the trigger to already be present on the job. Existing
tests all pass, and the new test failed without the changes to
WorkflowJob and PipelineTriggersJobProperty. Phew.

Note - I'm working on getting rid of the need for posting hardcoded
JSON and just using form manipulation.
```

Creating PR to get artifact for testing.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/workflow-job-plugin/28)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -445,6 +445,10 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements B
         return triggerProp;
     }
 
+    public void addTriggersJobPropertyWithoutStart(PipelineTriggersJobProperty prop) throws IOException {
+        super.addProperty(prop);
+    }
+
     public void setTriggers(List<Trigger<?>> inputTriggers) throws IOException {
         triggers = null;
         BulkChange bc = new BulkChange(this);
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -96,6 +96,7 @@ import org.jenkinsci.plugins.workflow.job.properties.DisableConcurrentBuildsJobP
 import org.jenkinsci.plugins.workflow.job.properties.PipelineTriggersJobProperty;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
@@ -445,6 +446,7 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements B
         return triggerProp;
     }
 
+    @Restricted(NoExternalUse.class)
     public void addTriggersJobPropertyWithoutStart(PipelineTriggersJobProperty prop) throws IOException {
         super.addProperty(prop);
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
@@ -145,14 +145,17 @@ public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
             throw new Descriptor.FormException(e, null);
         }
 
-        this.stopTriggers();
+        try {
+            owner.removeProperty(this);
+            PipelineTriggersJobProperty thisProp = new PipelineTriggersJobProperty(new ArrayList<Trigger>(trigList.toList()));
 
-        PipelineTriggersJobProperty thisProp = new PipelineTriggersJobProperty(new ArrayList<Trigger>(trigList.toList()));
-        thisProp.setOwner(owner);
+            owner.addTriggersJobPropertyWithoutStart(thisProp);
 
-        thisProp.startTriggers(true);
-
-        return thisProp;
+            thisProp.startTriggers(true);
+            return thisProp;
+        } catch (IOException e) {
+            return null;
+        }
     }
 
     @Extension(ordinal = -100)
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
@@ -49,9 +49,13 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 @SuppressWarnings("rawtypes")
 public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
+    private static final Logger LOGGER = Logger.getLogger(PipelineTriggersJobProperty.class.getName());
+
     private List<Trigger<?>> triggers = new ArrayList<>();
 
     @DataBoundConstructor
@@ -133,7 +137,7 @@ public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
     @CheckForNull
     @Override
     public PipelineTriggersJobProperty reconfigure(@Nonnull StaplerRequest req, @CheckForNull JSONObject form) throws Descriptor.FormException {
-        DescribableList<Trigger<?>,TriggerDescriptor> trigList = new DescribableList<>(Saveable.NOOP);
+        DescribableList<Trigger<?>, TriggerDescriptor> trigList = new DescribableList<>(Saveable.NOOP);
         try {
             JSONObject triggerSection = new JSONObject();
             if (form != null) {
@@ -145,6 +149,8 @@ public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
             throw new Descriptor.FormException(e, null);
         }
 
+        PipelineTriggersJobProperty oldProp = owner.getTriggersJobProperty();
+
         try {
             owner.removeProperty(this);
             PipelineTriggersJobProperty thisProp = new PipelineTriggersJobProperty(new ArrayList<Trigger>(trigList.toList()));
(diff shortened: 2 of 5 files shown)


KostyaSha:
@abayer is there any progress on it?

abayer:
cc @reviewbybees esp @jglick since this is actually my change in the first place. =)

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -145,14 +145,17 @@ public PipelineTriggersJobProperty reconfigure(@Nonnull StaplerRequest req, @Che
...
+            PipelineTriggersJobProperty thisProp = new PipelineTriggersJobProperty(new ArrayList<Trigger>(trigList.toList()));
 
-        thisProp.startTriggers(true);
+            owner.addTriggersJobPropertyWithoutStart(thisProp);
 
-        return thisProp;
+            thisProp.startTriggers(true);
+            return thisProp;
+        } catch (IOException e) {
+            return null;
```
oleg-nenashev: 🐛 Must be Error logging at least

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -445,6 +445,10 @@ public PipelineTriggersJobProperty getTriggersJobProperty() {
         return triggerProp;
     }
 
+    public void addTriggersJobPropertyWithoutStart(PipelineTriggersJobProperty prop) throws IOException {
```
jglick:
```java
@Restricted(NoExternalUse.class)
```

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
jglick: Should log.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -145,14 +145,17 @@ public PipelineTriggersJobProperty reconfigure(@Nonnull StaplerRequest req, @Che
             throw new Descriptor.FormException(e, null);
         }
 
-        this.stopTriggers();
-
-        PipelineTriggersJobProperty thisProp = new PipelineTriggersJobProperty(new ArrayList<Trigger>(trigList.toList()));
-        thisProp.setOwner(owner);
+        try {
+            owner.removeProperty(this);
```
oleg-nenashev: 🐛 The property gets removed, but it won't be restored if exception happens at the <code> addTriggersJobPropertyWithoutStart</code> stage

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -145,14 +145,17 @@ public PipelineTriggersJobProperty reconfigure(@Nonnull StaplerRequest req, @Che
             throw new Descriptor.FormException(e, null);
         }
 
-        this.stopTriggers();
```
jglick: Where is the `stop` coming from in this case? `removeProperty`?

src/test/java/org/jenkinsci/plugins/workflow/job/properties/QueryingMockTrigger.java
```diff
@@ -0,0 +1,86 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
```
jglick: Presumably you could pick the copyright holder (IANAL).

src/test/java/org/jenkinsci/plugins/workflow/job/properties/QueryingMockTrigger.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+
+public class QueryingMockTrigger extends Trigger<BuildableItem> {
+
+    /** elements are true or false for {@code start}, null for {@code stop} */
+    public static List<Boolean> startsAndStops = new ArrayList<>();
+
+    public transient boolean isStarted;
+
+    public transient boolean foundSelf = false;
```
jglick: Could this not be added more simply to the existing `MockTrigger`?

src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -167,6 +170,30 @@ public void configRoundTrip() throws Exception {
...
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "triggerPresent");
+        r.configRoundtrip(p);
+        assertNull(getTriggerFromList(QueryingMockTrigger.class,
+                p.getTriggersJobProperty().getTriggers()));
+        JenkinsRule.WebClient wc = r.createWebClient();
+        String newConfig = org.apache.commons.io.IOUtils.toString(
+                PipelineTriggersJobPropertyTest.class.getResourceAsStream(
+                        "/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest/triggerPresentDuringStart.json"), "UTF-8");
+        WebRequest request = new WebRequest(new URL(p.getAbsoluteUrl() + "configSubmit"), HttpMethod.POST);
+        wc.addCrumb(request);
```
jglick: I think you can more simply `createCrumbedUrl`.

KostyaSha: I thought it possible to use HtmlForm for configuration, but not clear what json is really changing.

src/test/java/org/jenkinsci/plugins/workflow/job/properties/QueryingMockTrigger.java
abayer: Well, I actually am the one who wrote the code in this PR. @KostyaSha just opened it from my repo. =)

KostyaSha: @abayer was in travelling and i found his changes good, so opened PR for review. Now i can't even review this PR because i become author.

src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
abayer: `createCrumbedUrl` didn't work - it adds a `GET` parameter, which made it impossible to `POST` the JSON as well. Yay?

I tried to go through the `HtmlForm` to get the end result and just couldn't get it working, hence `POST`ing the JSON directly.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
abayer: Will do.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Will do.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
abayer: Will fix.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
abayer: Will do.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
abayer: @jglick Yes, `removeProperty`.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
abayer: Ok, actually, not sure how we'd do this...

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -154,8 +160,19 @@ public PipelineTriggersJobProperty reconfigure(@Nonnull StaplerRequest req, @Che
             thisProp.startTriggers(true);
             return thisProp;
         } catch (IOException e) {
-            return null;
+            LOGGER.log(Level.WARNING, "could not configure triggers", e);
```
KostyaSha: Why not error?

KostyaSha: Should start from capital letter?

abayer: Was copy-pasting from https://github.com/jenkinsci/workflow-job-plugin/blob/95a1c862feb7ab5e1a765e693842b7e1f084f023/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java =)

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
oleg-nenashev: Well, likely it can be done via BulkOperation with rollback/commit if they are handled correctly

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -145,14 +149,30 @@ public PipelineTriggersJobProperty reconfigure(@Nonnull StaplerRequest req, @Che
...
+        }
 
-        thisProp.startTriggers(true);
+        if (owner.getTriggersJobProperty() == null && oldProp != null) {
+            try {
+                owner.addTriggersJobPropertyWithoutStart(oldProp);
+                oldProp.startTriggers(true);
+            } catch (IOException e) {
+                LOGGER.log(Level.WARNING, "could not revert to original configured triggers", e);
```
oleg-nenashev: Maybe makes sense to say "You need to reconfigure job manually" and then throw it as FormException? Otherwise the issue does not get propagated. 🐜

KostyaSha: FormException will kill all configuration

KostyaSha:
How much time do you need to fix thing that break existing installations?

KostyaSha:
#29

KostyaSha:
That's all?

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
abayer: Yeah, I'm wary of throwing a `FormException` - I'd rather just revert and warn, but I'm not locked into that.

KostyaSha: Fixed in #29

PR #33, opened 2016-11-16
[FIXED JENKINS-33721] - Add contextual term/kill in side panel.

abayer:
[JENKINS-33721](https://issues.jenkins-ci.org/browse/JENKINS-33721)

Replaces #4. Still need to figure out if we need a different icon, though.

cc @reviewbybees esp @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -121,6 +121,14 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
 
     private static final Logger LOGGER = Logger.getLogger(WorkflowRun.class.getName());
 
+    private enum StopState {
+        TERM, KILL;
+
+        public String url() {
+            return this.name().toLowerCase();
+        }
+    }
+
     /** null until started, or after serious failures or hard kill */
     private @CheckForNull FlowExecution execution;
 
@@ -135,6 +143,10 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         }
     };
     private transient StreamBuildListener listener;
+
+    private transient boolean allowTerm;
+
+    private transient boolean allowKill;
 
     /**
      * Flag for whether or not the build has completed somehow.
@@ -260,7 +272,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                             LOGGER.log(Level.WARNING, null, x);
                         }
                         executor.recordCauseOfInterruption(WorkflowRun.this, listener);
-                        printLater("term", "Click here to forcibly terminate running steps");
+                        printLater(StopState.TERM, "Click here to forcibly terminate running steps");
                     }
                 });
             }
@@ -302,13 +314,21 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         return asynchronousExecution;
     }
 
-    private void printLater(final String url, final String message) {
+    private void printLater(final StopState state, final String message) {
         Timer.get().schedule(new Runnable() {
             @Override public void run() {
                 if (!isInProgress()) {
                     return;
                 }
-                listener.getLogger().println(POSTHyperlinkNote.encodeTo("/" + getUrl() + url, message));
+                switch (state) {
+                    case TERM:
+                        setAllowTerm(true);
+                        break;
+                    case KILL:
+                        setAllowKill(true);
+                        break;
+                }
+                listener.getLogger().println(POSTHyperlinkNote.encodeTo("/" + getUrl() + state.url(), message));
             }
         }, 15, TimeUnit.SECONDS);
     }
@@ -338,7 +358,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
             }
             @Override public void onFailure(Throwable t) {}
         });
-        printLater("kill", "Click here to forcibly kill entire build");
+        printLater(StopState.KILL, "Click here to forcibly kill entire build");
     }
 
     /** Immediately kills the build. */
@@ -369,6 +389,22 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         }
         EnvVars.resolve(env);
         return env;
+    }
+
+    public void setAllowTerm(boolean b) {
+        this.allowTerm = b;
+    }
+
+    public boolean hasAllowTerm() {
+        return allowTerm && !allowKill;
+    }
+
+    public void setAllowKill(boolean b) {
+        this.allowKill = b;
+    }
+
+    public boolean hasAllowKill() {
+        return allowKill;
     }
 
     @GuardedBy("completed")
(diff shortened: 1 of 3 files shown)


abayer:
ping @jglick

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -121,6 +121,14 @@
 
     private static final Logger LOGGER = Logger.getLogger(WorkflowRun.class.getName());
 
+    private enum StopState {
+        TERM, KILL;
+
+        public String url() {
+            return this.name().toLowerCase();
```
jglick: Must specify `Locale.ENGLISH` or Turkish and Azeri users will be left wondering why there is no `doKıll` method!

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -371,6 +391,22 @@ public void doKill() {
         return env;
     }
 
+    public void setAllowTerm(boolean b) {
```
jglick: Why is this `public`? Easier to just inline it.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -371,6 +391,22 @@ public void doKill() {
...
+        this.allowTerm = b;
+    }
+
+    public boolean hasAllowTerm() {
```
jglick:
```java
@Restricted(DoNotUse.class) // Jelly
```

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -371,6 +391,22 @@ public void doKill() {
...
+        return allowTerm && !allowKill;
```
jglick: No, you should be allowed to retry `term` even if `kill` has been offered. (As you can already do via hyperlinks.)

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
```diff
@@ -34,6 +34,30 @@
...
+                        <l:task icon="images/24x24/edit-delete.png"
+                                href="${buildUrl.baseUrl}/term"
+                                title="${%Abort Running Steps}"
+                                contextMenu="false"
+                                post="true"
+                                permission="${it.CANCEL}"
+                                requiresConfirmation="true"
+                                confirmationMessage="Really forcibly abort running steps?"/>
+                    </j:when>
+                    <j:when test="${it.hasAllowKill()}">
```
jglick: As above, both options should be offered, so you can retry `term` rather than resorting to `kill`.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
```diff
@@ -34,6 +34,30 @@
...
+                <j:choose>
+                    <j:when test="${it.hasAllowTerm()}">
+                        <l:task icon="images/24x24/edit-delete.png"
+                                href="${buildUrl.baseUrl}/term"
+                                title="${%Abort Running Steps}"
+                                contextMenu="false"
+                                post="true"
+                                permission="${it.CANCEL}"
+                                requiresConfirmation="true"
+                                confirmationMessage="Really forcibly abort running steps?"/>
```
jglick: Make localizable.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
```diff
@@ -34,6 +34,30 @@
...
+                    </j:when>
+                    <j:when test="${it.hasAllowKill()}">
+                        <l:task icon="images/24x24/edit-delete.png"
+                                href="${buildUrl.baseUrl}/kill"
+                                title="${%Force Abort of Entire Build}"
+                                contextMenu="false"
+                                post="true"
+                                permission="${it.CANCEL}"
+                                requiresConfirmation="true"
+                                confirmationMessage="Really forcibly abort entire build?"/>
```
jglick: Ditto.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
```diff
@@ -84,6 +82,51 @@
...
+                story.j.waitForMessage("Hard kill!", b);
+                story.j.waitForCompletion(b);
+                story.j.assertBuildStatus(Result.ABORTED, b);
+            }
+        });
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.getItemByFullName("p", WorkflowJob.class);
+                WorkflowRun b = p.getBuildByNumber(1);
+                assertFalse(b.isBuilding());
```
jglick: This second step is gratuitous.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
```diff
@@ -84,6 +82,51 @@
...
+                story.j.waitForMessage("[1] bwahaha FlowInterruptedException #2", b);
+                assertFalse(hasTermOrKillLink(b, "term"));
+                assertFalse(hasTermOrKillLink(b, "kill"));
+                story.j.waitForMessage("Click here to forcibly terminate running steps", b);
+                assertTrue(hasTermOrKillLink(b, "term"));
+                assertFalse(hasTermOrKillLink(b, "kill"));
+                b.doTerm();
+                story.j.waitForMessage("[2] undead", b);
+                story.j.waitForMessage("Click here to forcibly kill entire build", b);
+                assertFalse(hasTermOrKillLink(b, "term"));
```
jglick: As above, should be true.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
abayer: Earlier work had it called more - switching to inline.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
abayer: Okiedokie.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
```diff
@@ -34,6 +34,26 @@
             <p:console-link/>
             <l:task icon="images/24x24/notepad.png" href="${buildUrl.baseUrl}/configure" title="${h.hasPermission(it,it.UPDATE)?'%Edit Build Information':'%View Build Information'}"/>
             <st:include page="delete.jelly"/>
+            <j:if test="${it.building and it.hasAllowTerm()}">
```
jglick: The `isBuilding()` check could be moved into the Java code.

abayer: Good point! Doing so.

jubel-han:
Hi, @jglick when will release the new version for this ? I tried to manually build the plugin after merge this PR, but it failed, I'm stuck by this issue now, Thanks!

PR #35, opened 2017-01-31
[JENKINS-24141] Implement RunWithSCMMixIn.RunWithSCM

abayer:
[JENKINS-24141](https://issues.jenkins-ci.org/browse/JENKINS-24141)

Downstream of https://github.com/jenkinsci/jenkins/pull/2730

Upstream of https://github.com/jenkinsci/email-ext-plugin/pull/155

cc @reviewbybees esp @jglick

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <jenkins-core.version>2.58-20170502.171727-6</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
         <jenkins-war.version>2.58-20170502.171808-6</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-        <jenkins-test-harness.version>2.22-20170502.170034-3</jenkins-test-harness.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins-test-harness/pull/58 is merged and released -->
+        <jenkins-test-harness.version>2.22</jenkins-test-harness.version>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <java.level>8</java.level>
--- a/pom.xml
+++ b/pom.xml
@@ -64,9 +64,9 @@
     <properties>
         <jenkins-core.version>2.58-20170501.155158-2</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
         <jenkins-war.version>2.58-20170501.155220-2</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-        <jenkins-test-harness.version>2.19</jenkins-test-harness.version>
+        <jenkins-test-harness.version>2.21</jenkins-test-harness.version>
         <no-test-jar>false</no-test-jar>
-        <workflow-support-plugin.version>2.2</workflow-support-plugin.version>
+        <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
     </properties>
     <dependencies>
         <dependency>
@@ -82,13 +82,19 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>1.15</version>
+            <version>2.29</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>structs</artifactId>
+            <version>1.6</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-basic-steps</artifactId>
-            <version>1.15</version>
+            <version>2.4</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -101,13 +107,13 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-scm-step</artifactId>
-            <version>2.2</version>
+            <version>2.4</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-durable-task-step</artifactId>
-            <version>2.0</version>
+            <version>2.11</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -62,11 +62,12 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins-core.version>2.58-20170501.171905-3</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-        <jenkins-war.version>2.58-20170501.171927-3</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-core.version>2.58-20170502.144018-4</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-war.version>2.58-20170502.144037-4</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
         <jenkins-test-harness.version>2.21</jenkins-test-harness.version>
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
+        <java.level>8</java.level>
     </properties>
     <dependencies>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
+        <jenkins.version>2.44-JENKINS-24141-SNAPSHOT</jenkins.version>
         <no-test-jar>false</no-test-jar>
     </properties>
     <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -62,9 +62,9 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins-core.version>2.58-20170502.144018-4</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-        <jenkins-war.version>2.58-20170502.144037-4</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-        <jenkins-test-harness.version>2.21</jenkins-test-harness.version>
+        <jenkins-core.version>2.58-20170502.171727-6</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-war.version>2.58-20170502.171808-6</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-test-harness.version>2.22-20170502.170034-3</jenkins-test-harness.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins-test-harness/pull/58 is merged and released -->
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <java.level>8</java.level>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.27-20170501.155323-2</version> <!-- TODO: Switch to 2.27 once https://github.com/jenkinsci/plugin-pom/pull/59 is released -->
+        <version>2.27</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -62,9 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins-core.version>2.58-20170502.171727-6</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-        <jenkins-war.version>2.58-20170502.171808-6</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-        <jenkins-test-harness.version>2.22</jenkins-test-harness.version>
+        <jenkins-core.version>2.58-20170502.192524-8</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-war.version>2.58-20170502.192544-8</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
         <no-test-jar>false</no-test-jar>
         <workflow-support-plugin.version>2.14</workflow-support-plugin.version>
         <java.level>8</java.level>
--- a/pom.xml
+++ b/pom.xml
@@ -62,8 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins-core.version>2.58-20170501.155158-2</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
-        <jenkins-war.version>2.58-20170501.155220-2</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-core.version>2.58-20170501.171905-3</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-war.version>2.58-20170501.171927-3</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
         <jenkins-test-harness.version>2.21</jenkins-test-harness.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -62,7 +62,7 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
+        <jenkins.version>2.44-JENKINS-24141-SNAPSHOT</jenkins.version>
```
jglick: Recommend using timestamped snapshots FWIW.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -152,6 +160,21 @@ public String url() {
...
+     * This is a list of {@link User#getId() user ids} who made a change
+     * since the last non-broken build. Can be null (which should be
+     * treated like empty set), because of the compatibility.
+     *
+     * <p>
+     * This field is semi-final --- once set the value will never be modified.
+     *
+     * @since 1.137
+     */
+    private volatile Set<String> culprits;
```
jglick: Should this be `transient`?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -189,6 +212,37 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
...
+
+            @Exported
+            public synchronized List<ChangeLogSet<? extends ChangeLogSet.Entry>> getChangeSets() {
+                if (changeSets == null) {
+                    changeSets = new ArrayList<>();
+                    for (SCMCheckout co : checkouts(null)) {
+                        if (co.changelogFile != null && co.changelogFile.isFile()) {
+                            try {
+                                ChangeLogSet<? extends ChangeLogSet.Entry> changeLogSet =
+                                        co.scm.createChangeLogParser().parse(asRun(), getEffectiveBrowser(co.scm), co.changelogFile);
```
jglick: BTW `getEffectiveBrowser` can be reverted now that we are in trunk.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -631,6 +685,9 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
         } catch (Exception x) {
             LOGGER.log(Level.WARNING, "failed to save " + this + " or perform log rotation", x);
         }
+        if (culprits == null) {
+            culprits = ImmutableSortedSet.copyOf(getCulpritIds());
```
jglick: Do we really need to compute this eagerly? Most builds will not need it AFAIK.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
abayer: I don't think so - it's not `transient` over in `AbstractBuild` historically.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
abayer: So this is aping `AbstractBuild` - trying to keep `culprits` write-once.

pom.xml
abayer: timestamped instead of more specific version?

jglick:
This supersedes #19 right?

pom.xml
jglick: I suspect you will need https://github.com/jenkinsci/plugin-pom/pull/59 (`deploy`ed as `2.27-20170501.155323-2`).

pom.xml
```diff
@@ -62,7 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
+        <jenkins-core.version>2.58-20170501.155158-2</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-war.version>2.58-20170501.155220-2</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
```
jglick: From the build log:

```
[INFO] Downloaded: http://repo.jenkins-ci.org/public/org/jenkins-ci/main/jenkins-core/2.58-SNAPSHOT/jenkins-core-2.58-20170501.155158-2.jar (10112 KB at 6591.4 KB/sec)
[INFO] Downloading: http://repo.jenkins-ci.org/public/org/jenkins-ci/main/jenkins-war/2.58-SNAPSHOT/jenkins-war-2.58-20170501.155220-2-war-for-test.jar
```

Looks right to me.

abayer: Yup, is good, I just screwed up and ran it with Java 7 initially, which silently didn't run tests without saying why. Good times, good times.

abayer:
Looks like there are hanging tests. Digging in.

abayer:
Further investigation shows it's barfing at https://github.com/jenkinsci/workflow-job-plugin/blob/c2626e064de5473883ff9c82da0bc4f12402f8f3/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L705 - NPEing, specifically. Joy!

abayer:
That there was some dumb code. I don't know why it was like that in the first place, but it's fixed now. =)

pom.xml
```diff
@@ -62,10 +62,11 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
-        <jenkins-test-harness.version>2.19</jenkins-test-harness.version>
+        <jenkins-core.version>2.58-20170501.171905-3</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-war.version>2.58-20170501.171927-3</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
```
jglick: may as well

```xml
<java.level>8</java.level>
```

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: The top-level `if` switch in the upstream PR seems very odd to me, but I cannot put my finger on what looks wrong. Probably the mixing of a method to _compute_ culprits with a method to _retrieve_ culprits.

At any rate, you need to document upstream that this logic only works when the list of culprit IDs is fixed at the time of build completion—it apparently cannot be computed on demand.

abayer: Aaaaah. Did not think that through, and that's actually, IMO, a dealbreaker - there's no real value to culprits in Pipeline if you can't do it during the build. So I'll let this stew for the night and see if I've got a better idea tomorrow.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -647,6 +701,13 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
         } catch (Exception x) {
             LOGGER.log(Level.WARNING, "failed to save " + this + " or perform log rotation", x);
         }
+        if (culprits == null) {
+            HashSet<String> tempCulpritIds = new HashSet<>();
+            for (User u : getCulprits()) {
+                tempCulpritIds.add(u.getId());
+            }
+            culprits = ImmutableSortedSet.copyOf(tempCulpritIds);
```
jglick: So we are computing this only at build completion. What about a `mail` step in a `post` `failure` block? When the `result` is still `null`? It is not obvious that the culprits logic will work at all in any practical situation. Needs some integration tests.

abayer: Yeah, strongly agreed. Did not think that through far enough.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -193,6 +216,33 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
         return runMixIn;
     }
 
+    @SuppressWarnings("unchecked") // untypable
```
jglick: Oh?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -193,6 +216,33 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
...
+    @Override public WorkflowRun asRun() {
+        return WorkflowRun.this;
+    }
+
+    @Override
+    @Exported
+    public synchronized List<ChangeLogSet<? extends ChangeLogSet.Entry>> getChangeSets() {
```
jglick: Just restore to its original position and add `@Override`, for a cleaner diff.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -107,7 +107,7 @@
 
     @Test public void funnyParameters() throws Exception {
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
-        p.setDefinition(new CpsFlowDefinition("echo \"a.b=${binding['a.b']}\"", true));
+        p.setDefinition(new CpsFlowDefinition("echo \"a.b=${params['a.b']}\"", true));
```
jglick: If I have a moment I could extract this to a separate PR bumping the `workflow-cps` test dep to, IIRC, 2.18.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
abayer: 'k.

pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.25</version>
+        <version>2.27-20170501.155323-2</version> <!-- TODO: Switch to 2.27 once https://github.com/jenkinsci/plugin-pom/pull/59 is released -->
```
jglick: released

pom.xml
```diff
@@ -62,10 +62,12 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
-        <jenkins-test-harness.version>2.19</jenkins-test-harness.version>
+        <jenkins-core.version>2.58-20170502.171727-6</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-war.version>2.58-20170502.171808-6</jenkins-war.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
+        <jenkins-test-harness.version>2.22</jenkins-test-harness.version>
```
jglick: now default, delete

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -364,4 +364,61 @@ public void globalNodePropertiesInEnv() throws Exception {
...
+    @Test
+    @Issue("JENKINS-24141")
+    public void culprits() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("import org.jvnet.hudson.test.FakeChangeLogSCM\n" +
+                "semaphore 'waitFirst'\n" +
+                "def testScm = new FakeChangeLogSCM()\n" +
+                "testScm.addChange().withAuthor('alice')\n" +
+                "node {\n" +
+                "    ws {\n" +
```
jglick: `ws` gratuitous I think

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -364,4 +364,61 @@ public void globalNodePropertiesInEnv() throws Exception {
...
+        WorkflowRun b1 = p.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("waitFirst/1", b1);
+        assertTrue(b1.getCulpritIds().isEmpty());
+        SemaphoreStep.success("waitFirst/1", null);
+
+        SemaphoreStep.waitForStart("waitSecond/1", b1);
+        Set<String> origCulprits = b1.getCulpritIds();
+        assertNotNull(origCulprits);
+        assertEquals(1, origCulprits.size());
+        assertTrue(origCulprits.contains("alice"));
```
jglick:
```java
assertEquals(Collections.singleton("alice"), b1.getCulpritIds());
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -364,4 +364,61 @@ public void globalNodePropertiesInEnv() throws Exception {
...
+
+        SemaphoreStep.success("waitSecond/1", null);
+
+        SemaphoreStep.waitForStart("waitThird/1", b1);
+        Set<String> secondCulprits = b1.getCulpritIds();
+        assertNotNull(secondCulprits);
+        assertEquals(2, secondCulprits.size());
+        assertTrue(secondCulprits.contains("alice"));
+        assertTrue(secondCulprits.contains("bob"));
```
jglick:
```java
assertEquals(ImmutableSet.of("alice", "bob"), b1.getCulpritIds());
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -364,4 +364,61 @@ public void globalNodePropertiesInEnv() throws Exception {
...
+        SemaphoreStep.success("waitThird/1", b1);
+
+        r.assertBuildStatusSuccess(r.waitForCompletion(b1));
+
+        Set<String> thirdCulprits = b1.getCulpritIds();
+        assertNotNull(thirdCulprits);
+        assertEquals(3, thirdCulprits.size());
+        assertTrue(thirdCulprits.contains("alice"));
+        assertTrue(thirdCulprits.contains("bob"));
+        assertTrue(thirdCulprits.contains("charlie"));
```
jglick: etc.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -364,4 +364,61 @@ public void globalNodePropertiesInEnv() throws Exception {
...
+        SemaphoreStep.waitForStart("waitThird/1", b1);
+        Set<String> secondCulprits = b1.getCulpritIds();
+        assertNotNull(secondCulprits);
+        assertEquals(2, secondCulprits.size());
+        assertTrue(secondCulprits.contains("alice"));
+        assertTrue(secondCulprits.contains("bob"));
+
+        SemaphoreStep.success("waitThird/1", b1);
+
+        r.assertBuildStatusSuccess(r.waitForCompletion(b1));
```
jglick: Should you not test the behavior when there are some broken builds as well? (Suffices to call `SemaphoreStep.failure` with `AbortException` I suppose.)

abayer: Yup, I should. Will do.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -364,4 +366,57 @@ public void globalNodePropertiesInEnv() throws Exception {
...
     }
 
+    @Test
+    @Issue("JENKINS-24141")
+    public void culprits() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("import org.jvnet.hudson.test.FakeChangeLogSCM\n" +
+                "semaphore 'waitFirst'\n" +
+                "def testScm = new FakeChangeLogSCM()\n" +
+                "testScm.addChange().withAuthor('alice' + env.BUILD_NUMBER)\n" +
```
jglick: `env.` prefix unnecessary BTW, so could use something like

```groovy
withAuthor(/alice$BUILD_NUMBER/)
```

abayer: Okiedokie.

pom.xml
```diff
@@ -62,10 +62,11 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
-        <jenkins-test-harness.version>2.19</jenkins-test-harness.version>
+        <jenkins-core.version>2.58-20170502.192524-8</jenkins-core.version> <!-- TODO: Switch to release once https://github.com/jenkinsci/jenkins/pull/2730 is merged and released -->
```
oleg-nenashev: Needs update to 2.60

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -156,6 +160,21 @@ public String url() {
...
+     *
+     * <p>
+     * This is a list of {@link User#getId() user ids} who made a change
+     * since the last non-broken build. Can be null (which should be
+     * treated like empty set), because of the compatibility.
+     *
+     * <p>
+     * This field is semi-final --- once set the value will never be modified.
+     *
+     * @since 1.137
```
oleg-nenashev: WAT? 🐜

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -156,6 +160,21 @@ public String url() {
     private transient boolean allowKill;
 
     /**
+     * Cumulative list of people who contributed to the build problem.
+     *
+     * <p>
+     * This is a list of {@link User#getId() user ids} who made a change
+     * since the last non-broken build. Can be null (which should be
```
oleg-nenashev: Should be annotated then. Seems to be a copy-paste from an ancient codebase

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -737,10 +757,25 @@ public boolean hasntStartedYet() {
...
+            } else {
+                culprits = ImmutableSortedSet.copyOf(tempCulpritIds);
+            }
+        }
+        return culprits;
+    }
+
+    @Override
+    public boolean shouldCalculateCulprits() {
+        return isBuilding() || culprits == null;
```
oleg-nenashev: Is it affected by the issue in Jenkins core API, when `isBuilding()` effectively check only the last run and does not care about others? Maybe be a problem for parallel runs

jglick:
> core dependency needs to be updated before the merge

Obviously, but this is not a topic for code review, it is just a task for the plugin maintainer (myself in this case).

PR #41, opened 2017-03-09
[JENKINS-42446] Triggers should be started on job creation

scoheb:
[JENKINS-42446](https://issues.jenkins-ci.org/browse/JENKINS-42446)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -196,6 +196,8 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements B
         } else {
             quietPeriod = null;
         }
+        getTriggersJobProperty().stopTriggers();
+        getTriggersJobProperty().startTriggers(Items.currentlyUpdatingByXml());
     }
 
 
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
@@ -38,6 +38,7 @@ import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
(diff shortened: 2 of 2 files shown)


scoheb:
@jenkinsci/code-reviewers

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -196,6 +196,8 @@ public void setDefinition(FlowDefinition definition) {
         } else {
             quietPeriod = null;
         }
+        getTriggersJobProperty().stopTriggers();
+        getTriggersJobProperty().startTriggers(Items.currentlyUpdatingByXml());
```
jglick: Not really the right place to do this. `setOwner` seems more appropriate, but this was dropped in 09adb56241cbcac3baf2625f063af8da7ba87bee; perhaps @abayer recalls why.

src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -93,10 +94,10 @@ public void submitCallsStartTrue() throws Exception {
         t = (MockTrigger) p.getTriggers().get(t.getDescriptor());
         assertNotNull(t);
         // The  first "null, false" is due to the p.addTrigger(t) call and won't apply in the real world.
-        assertEquals("[null, false, null, true]", MockTrigger.startsAndStops.toString());
+        assertEquals("[null, false, null, true, null, false]", MockTrigger.startsAndStops.toString());
```
jglick: So it seems that we are now calling `stop()` and `start(false)` gratuitously after a form resubmission.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
abayer: Not a clue. The trigger stop/start stuff is black magic - I basically just worry about whether the tests pass.

daniel-beck:
Does this break `isHoldOffBuildUntilSave()`?

jglick:
@abayer so what about the redundant stop/start calls? Does that not look wrong?

src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
abayer: No one actually understands why or when we're calling those, honestly.

PR #42, opened 2017-03-22
Avoid adding a PipelineTriggersJobProperty unless some triggers were actually configured

jglick:
Incorporates #41, without which `triggerPresentDuringStart` would fail, though I do not think that is mergeable as written.

The idea of this PR is to ensure that in **Pipeline Syntax** for `properties`, if you do not select any triggers you will see just

```groovy
properties([])
```

and not

```
properties([pipelineTriggers([])])
```

@reviewbybees esp. @abayer


jglick:
IIRC the actual patch here is independent of #41, but it cannot be sanely tested without that.

jglick:
Not sure about the impact of #41, but anyway I have made this mergeable again @abayer.

PR #43, opened 2017-04-06
[FIXED JENKINS-43396] Global NodePropertys should be in environment

abayer:
[JENKINS-43396](https://issues.jenkins-ci.org/browse/JENKINS-43396)

They currently show up once you enter a node block, but not for things
outside node blocks. This fixes that by adding them to
WorkflowRun.getEnvironment(listener).

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -393,8 +393,11 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     @Override public EnvVars getEnvironment(TaskListener listener) throws IOException, InterruptedException {
         EnvVars env = super.getEnvironment(listener);
 
-        for (NodeProperty nodeProperty: Jenkins.getInstance().getGlobalNodeProperties()) {
-            nodeProperty.buildEnvVars(env,listener);
+        Jenkins instance = Jenkins.getInstance();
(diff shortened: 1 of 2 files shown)


daniel-beck:
How does this affect values referencing other variables at variously contributed extension point implementations?

Additionallyy it's reasonable to expect node variables to be able to reference e.g. `${EXECUTOR_NUMBER}`. Will this change break something like that _within node blocks_ when it wasn't broken before?

abayer:
@daniel-beck This is [pretty much lifted straight from `Computer`](https://github.com/jenkinsci/jenkins/blob/09cfe3bda60341edb07ade226e24196a3f875019/core/src/main/java/hudson/model/Computer.java#L1177-L1179) - it's just the `globalNodeProperties`, not the master-specific `nodeProperties`. Is it reasonable/feasible for global node properties to be doing things with `EXECUTOR_NUMBER`?

jglick:
> Is it reasonable/feasible for global node properties to be doing things with EXECUTOR_NUMBER?

I do not think so.

daniel-beck:
@jglick 

> I do not think so.

I used it to get one Gradle cache per executor to prevent lock contention with many small(ish) builds. `GRADLE_USER_HOME=/path/wherever/$EXECUTOR_NUMBER` worked like a charm. :smiley: 

Really, what other use cases for `EXECUTOR_NUMBER` are there?

jglick:
Such a trick makes sense during a Pipeline build using `withEnv`, not as a node property.

PR #44, opened 2017-04-25
Ensure WorkflowRun instances are collectible immediately after completion

jglick:
Perusing a heap dump, I noticed a number of `WorkflowRun` instances which did not exist on disk, and whose root references came from `WebAppMain$1.records`, i.e., `Jenkins.logRecords`, which keeps the last 256 messages logged at `INFO` or higher.

Besides being confusing, this wastes a bit of heap if there are not many other messages being logged and the builds would otherwise be either deleted, or released from `BuildReference.SoftHolder`.

`Run.execute` has the same issue, though an `AbstractBuild` probably has a smaller memory footprint, and this is what I was interested in. Could file a matching core PR if anyone cares.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.21</version>
+        <version>2.25</version>
         <relativePath />
     </parent>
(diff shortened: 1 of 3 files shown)


jglick:
Need to check the test failures here.

jglick:
No, the failure is reproducible; I need to analyze it.

PR #51, opened 2017-05-19
[JENKINS-43055] Support FlowExecutionListener

abayer:
[JENKINS-43055](https://issues.jenkins-ci.org/browse/JENKINS-43055)

Replaces https://github.com/jenkinsci/workflow-job-plugin/pull/50 and https://github.com/jenkinsci/workflow-cps-plugin/pull/127, downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/36

cc @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.14-20170519.135335-5</version> <!-- TODO: update when https://github.com/jenkinsci/workflow-api-plugin/pull/36 is merged and released -->
+            <version>2.14-20170519.151127-6</version> <!-- TODO: update when https://github.com/jenkinsci/workflow-api-plugin/pull/36 is merged and released -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.14-20170519.151127-6</version> <!-- TODO: update when https://github.com/jenkinsci/workflow-api-plugin/pull/36 is merged and released -->
+            <version>2.14</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
```diff
@@ -175,4 +184,96 @@ public DescriptorImpl() {
...
+                SemaphoreStep.success("wait/1", null);
+
+                story.j.assertBuildStatus(Result.FAILURE, story.j.waitForCompletion(b));
+                story.j.assertLogContains("Running for listener", b);
+
+                ExecListener listener = ExtensionList.lookup(FlowExecutionListener.class).get(ExecListener.class);
+                assertNotNull(listener);
+                assertEquals(1, listener.started);
+                assertEquals(1, listener.resumed);
+                assertEquals(1, listener.finished);
```
jglick: Could add an extra `semaphore` to demonstrate started & resumed & !finished.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
```diff
@@ -175,4 +184,96 @@ public DescriptorImpl() {
...
+                WorkflowRun b = p.getLastBuild();
+                assertTrue(b.isBuilding());
+                SemaphoreStep.success("wait/1", null);
+
+                story.j.assertBuildStatus(Result.FAILURE, story.j.waitForCompletion(b));
+                story.j.assertLogContains("Running for listener", b);
+
+                ExecListener listener = ExtensionList.lookup(FlowExecutionListener.class).get(ExecListener.class);
+                assertNotNull(listener);
+                assertEquals(1, listener.started);
```
jglick: Potentially confusing to readers. This `started` is not the same as the one in the first step: `onRunning(…, true)` increments _both_ fields, but we are using a new listener!

This is why I wonder whether having a single method overloaded to two events is wise.

abayer: Yeaaaah, it's not. Fixing.

abayer:
:bee: Hold off on this for the moment - something's wrong in terms of the `SyntheticStageGraphListener` attaching in Declarative.

abayer:
Ok, findbugs fixed - @jglick, back to you for final review/merging/2.11.1 release.

pom.xml
```diff
@@ -77,6 +77,11 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-api</artifactId>
+            <version>2.15</version>
```
jglick:
```
java.io.IOException: Pipeline: API v2.15 failed to load.
 - Pipeline: Step API v2.9 is older than required. To fix, install v2.10 or later.
```

abayer: Fixing.

PR #54, opened 2017-05-26
[FIXED JENKINS-44520] - Invoke LaunchDecorators for node when polling with workspace

oleg-nenashev:
Just noticed it during the investigation of unrelated issue. The test may be overcomplicated, but it adds some extra coverage to the plugin codebase.

https://issues.jenkins-ci.org/browse/JENKINS-44520

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
@@ -4,12 +4,34 @@ import com.gargoylesoftware.htmlunit.HttpMethod;
 import com.gargoylesoftware.htmlunit.WebRequest;
 import com.gargoylesoftware.htmlunit.html.HtmlCheckBoxInput;
 import com.gargoylesoftware.htmlunit.html.HtmlForm;
+import hudson.EnvVars;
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
CC @jglick

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
```diff
@@ -99,4 +123,79 @@ public void addAction() throws Exception {
...
+                public String getUrlName() {
+                    return "foo";
+                }  
+            };
+        }
+
+        @Override
+        public PollingResult compareRemoteRevisionWith(Job<?, ?> project, Launcher launcher, FilePath workspace, TaskListener listener, SCMRevisionState baseline) throws IOException, InterruptedException {
+            List<String> commandLine = 
+                    Arrays.asList("echo", hudson.remoting.Launcher.isWindows() ? "INJECTED=%INJECTED%" : "INJECTED=$INJECTED");
```
jglick: fails on Windows: `echo` is not a command, it is a `cmd.exe` built-in I suppose

oleg-nenashev:
I am no longer interested in it

PR #57, opened 2017-06-21
Remove unnecessary stops/starts of triggers

scoheb:
In the case where you are using the **properties()** step with job properties other that the **PipelineTriggersJobProperty**, we do not want
stops and starts of a trigger every time we update a non-PipelineTriggersJobProperty.

This can pose a problem for a trigger that may have established a connection to an external system and it is forced to stop/start
frequently.


scoheb:
It would be great that, if this is merged, it gets ported to branches **2.11.x** and **2.12.x**

svanoort:
I'm going to merge once the CI comes up clean and then cut a release with this in it.

PR #58, opened 2017-06-22
[JENKINS-45067] - Prevent NPE in PipelineTriggersJobProperty#startTriggers() when owner is not assigned

oleg-nenashev:
When somebody creates a property in API and occasionally triggers the startTriggers() method, Trigger#start(owner, bool) gets invoked with null argument, which violates the contract and causes NPE. I think this NPE should be prevented, because such RunTime exception is not documented in the class Javadoc, and hence it may cause issues in the API user code.

https://issues.jenkins-ci.org/browse/JENKINS-45067

@reviewbybees @abayer @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
@@ -98,7 +98,21 @@ public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
         }
     }
 
+    /**
+     * Starts all triggers.
+     * This method requires the property instance to be added to the {@link WorkflowJob} instance.
+     * Otherwise the triggers will be just skipped.
+     * 
+     * @param newInstance {@code true} if it is a newly created instance.
+     */
     public void startTriggers(boolean newInstance) {
+        if (owner == null && LOGGER.isLoggable(Level.FINE)) {
+            // This is a normal behavior. Even if we miss the triggers,
+            // They will be started once WorkflowJob#addProperty() is called (standard method override).
+            LOGGER.log(Level.FINE, "Cannot start triggers. Owner of the JobProperty has not been assigned yet", 
+                    new IllegalStateException("JobProperty owner is null"));
+        }
+        
         for (Trigger trigger : triggers) {
             try {
                 trigger.start(owner, newInstance);
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
@@ -106,11 +106,14 @@ public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
      * @param newInstance {@code true} if it is a newly created instance.
      */
     public void startTriggers(boolean newInstance) {
-        if (owner == null && LOGGER.isLoggable(Level.FINE)) {
-            // This is a normal behavior. Even if we miss the triggers,
-            // They will be started once WorkflowJob#addProperty() is called (standard method override).
-            LOGGER.log(Level.FINE, "Cannot start triggers. Owner of the JobProperty has not been assigned yet", 
-                    new IllegalStateException("JobProperty owner is null"));
+        if (owner == null) {
+            if (LOGGER.isLoggable(Level.FINE)) {
+                // This is a normal behavior. Even if we miss the triggers,
+                // They will be started once WorkflowJob#addProperty() is called (standard method override).
+                LOGGER.log(Level.FINE, "Cannot start triggers. Owner of the JobProperty has not been assigned yet", 
+                        new IllegalStateException("JobProperty owner is null"));
+            }
+            return;
         }
         
         for (Trigger trigger : triggers) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
@@ -110,7 +110,7 @@ public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
             if (LOGGER.isLoggable(Level.FINE)) {
                 // This is a normal behavior. Even if we miss the triggers,
                 // They will be started once WorkflowJob#addProperty() is called (standard method override).
-                LOGGER.log(Level.FINE, "Cannot start triggers. Owner of the JobProperty has not been assigned yet", 
+                LOGGER.log(Level.WARNING, "Cannot start triggers. Owner of the JobProperty has not been assigned yet", 
                         new IllegalStateException("JobProperty owner is null"));
             }
             return;
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
@@ -30,6 +30,7 @@ import com.gargoylesoftware.htmlunit.WebRequest;
 import com.gargoylesoftware.htmlunit.util.NameValuePair;
 import hudson.model.Item;
 import hudson.model.Items;
+import hudson.model.Job;
 import hudson.triggers.TimerTrigger;
 import hudson.triggers.Trigger;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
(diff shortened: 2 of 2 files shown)


jglick:
I would rather start with fixing up #41, which might solve the issue (as well as unblocking #42). Probably needs me or @abayer to do a deep-dive.

oleg-nenashev:
Yes, from the JENKINS-42446 PoV it is rather a hack. Though it still can be considered as a valid fix for the public API

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -98,7 +98,21 @@ public void stopTriggers() {
...
+     * @param newInstance {@code true} if it is a newly created instance.
+     */
     public void startTriggers(boolean newInstance) {
+        if (owner == null && LOGGER.isLoggable(Level.FINE)) {
+            // This is a normal behavior. Even if we miss the triggers,
+            // They will be started once WorkflowJob#addProperty() is called (standard method override).
+            LOGGER.log(Level.FINE, "Cannot start triggers. Owner of the JobProperty has not been assigned yet", 
+                    new IllegalStateException("JobProperty owner is null"));
+        }
+
```
jglick: …but you are continuing? I do not understand what the purpose of this PR is.

oleg-nenashev: oops

recampbell:
Jesse suggests we put this fix on hold.

oleg-nenashev:
@jglick 
> Being defensive is fine, but ought to log a real warning.

So what would be your proposal? IMHO I have put a real warning for the case with null warning

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -98,7 +98,24 @@ public void stopTriggers() {
...
+     * Starts all triggers.
+     * This method requires the property instance to be added to the {@link WorkflowJob} instance.
+     * Otherwise the triggers will be just skipped.
+     * 
+     * @param newInstance {@code true} if it is a newly created instance.
+     */
     public void startTriggers(boolean newInstance) {
+        if (owner == null) {
+            if (LOGGER.isLoggable(Level.FINE)) {
+                // This is a normal behavior. Even if we miss the triggers,
```
jglick: Not a normal behavior.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -98,7 +98,24 @@ public void stopTriggers() {
...
+    /**
+     * Starts all triggers.
+     * This method requires the property instance to be added to the {@link WorkflowJob} instance.
+     * Otherwise the triggers will be just skipped.
+     * 
+     * @param newInstance {@code true} if it is a newly created instance.
+     */
     public void startTriggers(boolean newInstance) {
+        if (owner == null) {
+            if (LOGGER.isLoggable(Level.FINE)) {
```
jglick: Should be at `WARNING`.

src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -193,6 +196,37 @@ public void triggerPresentDuringStart() throws Exception {
...
+    @Issue("JENKINS-45067")
+    public void triggerMethodsShouldNotThrowNPEWhenNotAssigned() {
+        ProjectDependentTrigger t = new ProjectDependentTrigger();
+        PipelineTriggersJobProperty prop = new PipelineTriggersJobProperty(Arrays.asList(t));
+        
+        prop.startTriggers(true);
+        assertFalse("Trigger#start() API requires project to be non-null, hence it must not be invoked", t.triedToStart);
+        
+        prop.stopTriggers();
+        assertTrue("Trigger#stop() API does not require project to be non-null, hence it should be invoked", t.triedToStop);
```
jglick: Could also use `LoggerRule` to verify that a warning was issued.

src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -193,6 +196,37 @@ public void triggerPresentDuringStart() throws Exception {
...
+    }
+    
+    private static class ProjectDependentTrigger extends Trigger<Job<?, ?>> {
+
+        boolean triedToStart, triedToStop;
+        
+        @Override
+        public void start(Job<?, ?> project, boolean newInstance) {
+            triedToStart = true;
+            System.out.println("Triggering " + project.getFullDisplayName());
```
KostyaSha: system out?

oleg-nenashev:
@reviewbybees @abayer @svanoort @jglick Does it make sense to keep it open? The underlying issue has been fixed, and IIUC @jglick disagrees with the defensive code proposed here. I can finalize it if there is interest from the current plugin maintainers, but otherwise I would just close it.

svanoort:
1 vote for, 1 vote against... what do we do here, tie-breakers?

oleg-nenashev:
I suppose we summon @abayer and @jenkinsci/code-reviewers

oleg-nenashev:
@svanoort so I leave the decision to you.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
bitwiseman:
```suggestion
            if (LOGGER.isLoggable(Level.WARNING)) {
```

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -98,7 +98,24 @@ public void stopTriggers() {
...
+                // This is a normal behavior. Even if we miss the triggers,
+                // They will be started once WorkflowJob#addProperty() is called (standard method override).
+                LOGGER.log(Level.FINE, "Cannot start triggers. Owner of the JobProperty has not been assigned yet", 
+                        new IllegalStateException("JobProperty owner is null"));
```
bitwiseman: @oleg-nenashev 
My thinking here is that generally if a method does not succeed it should throw an exception, not quietly do nothing.  At the same time, null is a possible value and NPEs are bad.  

How about throwing  `IllegalStateException` instead of just logging?

src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
bitwiseman: Use `LOGGER` instead?

bitwiseman:
@oleg-nenashev 
It's been a while and Sam obviously didn't marshal this forward.  It is your call if you want to close or update this.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
jglick: My very dim recollection of the situation was that we occasionally get null values due to mistakes elsewhere, or maybe race conditions, so it would be best to log a warning and then move on.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -98,7 +98,24 @@ public void stopTriggers() {
...
+     * Otherwise the triggers will be just skipped.
+     * 
+     * @param newInstance {@code true} if it is a newly created instance.
+     */
     public void startTriggers(boolean newInstance) {
+        if (owner == null) {
+            if (LOGGER.isLoggable(Level.FINE)) {
+                // This is a normal behavior. Even if we miss the triggers,
+                // They will be started once WorkflowJob#addProperty() is called (standard method override).
+                LOGGER.log(Level.FINE, "Cannot start triggers. Owner of the JobProperty has not been assigned yet",
```
bitwiseman:
```suggestion
                LOGGER.log(Level.WARNING, "Cannot start triggers. Owner of the JobProperty has not been assigned yet", 
```

src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
bitwiseman:
```suggestion
                // This happens occasionally. Even if we miss the triggers,
```

oleg-nenashev:
@bitwiseman throwing an informative exception would be a step forward anyway. Better than NPE

jglick:
Unclear what the root cause is here, or whether the problem still occurs, and the PR does not really assist in diagnosis.

PR #59, opened 2017-07-27
[JENKINS-44548] Extend robustness fix to any runtime exception from FlowExecution.onLoad

jglick:
[JENKINS-44548](https://issues.jenkins-ci.org/browse/JENKINS-44548)

Second line of defense after https://github.com/jenkinsci/workflow-cps-plugin/pull/158: even if we did get an NPE, at least do not throw it out of `WorkflowRun.onLoad`.

@reviewbybees


jglick:
@reviewbybees done

PR #60, opened 2017-08-01
2.11.x: Add copyCheckouts(WorkflowRun) method.

abayer:
happening.

cc @reviewbybees


jglick:
Just do review in #62, and if and when merged, it can be backported (`git cherry-pick -x -m1 $merge_commit`) without additional review (assuming the backport is clean).

abayer:
@jglick Gotcha, closing.

PR #61, opened 2017-08-01
2.12.x: Add copyCheckouts(WorkflowRun) method.

abayer:
happening.

cc @reviewbybees


jglick:
Do review in #62; this could be closed.

PR #62, opened 2017-08-01
master: Add copyCheckouts(WorkflowRun) method.

abayer:
happening.

cc @reviewbybees


abayer:
@jglick Downstream PR is coming but is CloudBees-internal. I thought about `FlowCopier`, but I wasn't sure what else might be calling that and wasn't sure if we wanted to be copying the checkouts for everything that could be using `FlowCopier`.

PR #64, opened 2017-08-03
[FIXED JENKINS-45460] Allow adding triggers after migration.

abayer:
[JENKINS-45460](https://issues.jenkins-ci.org/browse/JENKINS-45460)

Before this, if you had triggers configured before upgrading from a
version that didn't yet have PipelineTriggersJobProperty,
PipelineTriggersJobProperty.setTriggers(List) was called via
WorkflowJob.onLoad with the old DescribableList triggers.toList() -
which returns an UnmodifiableList. Oops.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
@@ -71,7 +71,7 @@ public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
     }
 
     public void setTriggers(List<Trigger<?>> triggers) {
-        this.triggers = triggers;
+        this.triggers.addAll(triggers);
     }
 
     public List<Trigger<?>> getTriggers() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
@@ -71,7 +71,7 @@ public class PipelineTriggersJobProperty extends JobProperty<WorkflowJob> {
     }
 
     public void setTriggers(List<Trigger<?>> triggers) {
-        this.triggers.addAll(triggers);
+        this.triggers = new ArrayList<>(triggers);
     }
 
     public List<Trigger<?>> getTriggers() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobProperty.java
```diff
@@ -71,7 +71,7 @@ public PipelineTriggersJobProperty(List<Trigger> triggers) {
     }
 
     public void setTriggers(List<Trigger<?>> triggers) {
-        this.triggers = triggers;
+        this.triggers.addAll(triggers);
```
jglick: This is rather surprising. I would not expect direct API calls, but they would work weirdly if so.

Would it not be simpler to replace the `Collections.unmodifiableList()`, wherever that is, with `new ArrayList<>()`?

Or to make this code be

```java
this.triggers = new ArrayList<>(triggers);
```

abayer: Works out the same, I think - we init `this.triggers` anyway. But I can change it here if you'd like.

The `unmodifiableList` is off in [`PersistedList#toList`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/util/PersistedList.java#L190)'s call to ['CopyOnWriteList#getView`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/util/CopyOnWriteList.java#L145). I tend to think that probably should stay as is regardless.

abayer: (and I've changed to your suggestion)

src/test/java/org/jenkinsci/plugins/workflow/job/properties/PipelineTriggersJobPropertyTest.java
```diff
@@ -141,6 +141,14 @@ public void triggerMigration() throws Exception {
 
         assertNotNull(((MockTrigger)mockFromProp).currentStatus());
         assertEquals("[null, false, null, false, null, false]", MockTrigger.startsAndStops.toString());
+
+        // Verify that we can replace an existing trigger.
+        Trigger newTimerTrigger = new TimerTrigger("@hourly");
+
+        p.addTrigger(newTimerTrigger);
+
+        Trigger newTimerFromJob = p.getTriggers().get(newTimerTrigger.getDescriptor());
```
jglick: Maybe also verify that there is exactly one `TimerTrigger`?

abayer: Good point.

PR #66, opened 2017-08-24
Allow pipeline output decoration

lucamilanesio:
Allow other plugins to provide a decorator to filter
the output of a Jenkins Pipeline.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -71,6 +71,7 @@ import hudson.util.DescribableList;
 import hudson.widgets.HistoryWidget;
 import java.io.File;
 import java.io.IOException;
+import java.io.OutputStream;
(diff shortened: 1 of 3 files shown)


svanoort:
Seems reasonable to me @lucamilanesio but do you have an implementation taking advantage of the new API to go along with it?  Generally we like to have at least one consumer for net-new APIs.

lucamilanesio:
Yes, will push shorty a plug that extends the current logstash plugin for streaming the pipeline output, will showcase as well on Tuesday at JenkinsWorld

svanoort:
Okay, this is probably something that will be of special interest to @jglick since he's done work specifically related to external logging.

lucamilanesio:
@jglick any feedback? The logger decoration is very useful with the current master and with the additional refactoring that you are currently doing on your PR.

jglick:
Anyway should be obsoleted by #105.

bitwiseman:
@lucamilanesio  Would you agree this obsolete?  Should this PR be closed?

PR #68, opened 2017-08-31
[JENKINS-37845] Return SCM of CpsScmFlowDefinition if there is no pri…

silbernm:
This pull request significantly improves the situation for the issues
* https://issues.jenkins-ci.org/browse/JENKINS-35132
* https://issues.jenkins-ci.org/browse/JENKINS-37217
* https://issues.jenkins-ci.org/browse/JENKINS-37845
* https://issues.jenkins-ci.org/browse/JENKINS-45120

The issue is that a pipeline job which has not yet been built can not be triggered by a Github push trigger. Github push triggering is only available after at least one manual build.

The fix solves the issue for the typical use-case where the Jenkinsfile is contained in the same Github repository which hosts the source to be built. In that use-case, the pipeline job is properly triggered on a Github push, even if the pipeline job has *not* been triggered manually before.

The fix does not solve the issue completely for use-cases
* where multiple Github repositories are used, or
* where the Jenkinsfile and source code are located in different Github repositories.

In these use-cases, a yet unbuilt pipeline job is only triggered by a Github push to the repository containing the Jenkinsfile. Still, this is an improvement over the current situation. I don't see any straight-forward path to solve the issue in these use-cases completely.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,7 +89,6 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
             <version>2.29</version>
-            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -71,6 +71,7 @@ import hudson.util.DescribableList;
 import hudson.widgets.HistoryWidget;
 import java.io.File;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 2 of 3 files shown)


silbernm:
@jenkinsci/code-reviewers: I was not able to find infos on how to collaborate, but I guess I need to send this notification :-)

pom.xml
```diff
@@ -89,7 +89,6 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
             <version>2.29</version>
-            <scope>test</scope>
```
jglick: No.

abayer:
If this is truly worth doing, I'd suggest adding `getScm()` to `FlowDefinition` in `workflow-api`, returning `null` by default (with `CpsScmFlowDefinition` overriding that). Then we don't need to do circular dependencies.

silbernm:
Hi @abayer, I don't think this is truly worth doing anymore, especially if it requires an API change.

We are now working with the multibranch pipeline, and we trigger a repository scan on job creation. This solves the problem for us.

PR #70, opened 2017-09-18
[JENKINS-46934] Avoid blocking forever due to log rotation

jglick:
[JENKINS-46934](https://issues.jenkins-ci.org/browse/JENKINS-46934)

* Call `logRotate` asynchronously.
* Impose a timeout on `Owner.get`.
* While waiting for `Owner.get`, record the build in the thread name.

@reviewbybees


jglick:
There is timeout on log rotation here. Perhaps you misread the patch.

abayer:
@jglick Durr. Mornings, man.

jglick:
At any rate I suppose this is @reviewbybees done unless you change your mind.

abayer:
Yup, was waiting to see if anyone else had any input. But they don't, so merging!

PR #72, opened 2017-09-21
[JENKINS-46945] Run duration set before RunListener.fireCompleted

justinharringa:
Doing this will put the calculation of duration at the same location
where this is handled in core Jenkins' Run class:
https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/Run.java#L1758

Details here: https://issues.jenkins-ci.org/browse/JENKINS-46945

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -657,6 +657,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     /** Handles normal build completion (including errors) but also handles the case that the flow did not even start correctly, for example due to an error in {@link FlowExecution#start}. */
     private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
         setResult(r);
+        duration = Math.max(0, System.currentTimeMillis() - getStartTimeInMillis());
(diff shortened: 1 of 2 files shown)


justinharringa:
Any thoughts on this @jglick or @abayer? Sorry to ping you on this.

justinharringa:
@jglick I added a real-world scenario and can put together a test for it. Is there an easy way to inject a RunListener into Jenkins to validate this? I think I have some ideas but there might be an easier way. Thanks!

justinharringa:
Hey @jglick, good point on the test. I've added a test in 5a468ee and updated the JIRA a bit. Hopefully the JIRA is a bit more specific and the test helps. The test breaks if the change in WorkflowRun is reverted.

justinharringa:
Sorry to bug you again on this @jglick. Does this seem reasonable? This PR gives us better metrics and a consistent interface to get build time metrics. 

Could someone kick off the build again, it seems like there were some environmental errors at the time of the build that seem completely unrelated to my changes.

justinharringa:
Any additional thoughts on this @jglick? I had updated the ticket and added tests.

jglick:
> Sorry to ping you on this.

BTW use the “request a review” gesture when possible. Otherwise this message goes into the 6164 and counting unread messages from GitHub.

justinharringa:
Will do. Sorry about that.

svanoort:
Thank you for the contribution -- I'm going ahead and merging this now.

PR #67, opened 2017-08-24
[JENKINS-29537] Take advantage of Run.getEnvironment calling ParametersAction.buildEnvironment

jglick:
Downstream of https://github.com/jenkinsci/jenkins/pull/2993.

@reviewbybees

PR #74, opened 2017-09-27
Using WithThreadName

jglick:
Downstream of https://github.com/jenkinsci/workflow-support-plugin/pull/46. Picks up a couple updates from #67.

@reviewbybees


jglick:
@reviewbybees done

abayer:
Will merge once `workflow-support` 2.16 is out and this is updated accordingly.

svanoort:
workflow-support 2.16 is released so this is mergeable once the dependency is updated

PR #75, opened 2017-09-28
Provide job property for durability hints & add ability to disable pipeline resume [JENKINS-33761]

svanoort:
[https://issues.jenkins-ci.org/browse/JENKINS-33761](https://issues.jenkins-ci.org/browse/JENKINS-33761)

TODO:

- [x] Attempt to kill non-resumable pipeline upon Jenkins terminating, cleanly
- [x] If non-resumable pipeline didn't kill cleanup at shutdown, terminate with extreme prejudice at startup.
- [x] Manual testing with multibranch and the properties() step
- [x] Offer hints to lower-level APIs that we don't care about resuming after restart so we can simplify our operations (may need a workflow-api change too).


svanoort:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/55 ultimately to let us do some tracking of durability rules. 

We need *somehow* to set the durability to make no guarantees if resume is off.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/DisableResumeJobProperty.java
```diff
@@ -0,0 +1,26 @@
+package org.jenkinsci.plugins.workflow.job.properties;
+
+import hudson.Extension;
+import jenkins.model.OptionalJobProperty;
+import org.jenkinsci.Symbol;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+/**
+ * Disables resuming a pipeline if the master restarts - the run will simply fail instead, just like a FreeStyle job.
```
jglick: Should then be paired with a change to `workflow-durable-task-step` to make `sh`/etc. use synchronous mode.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -258,9 +264,33 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
             if (definition == null) {
                 throw new AbortException("No flow definition, cannot run");
             }
+
+            boolean loggedHintOverride = false;
+            if (getParent().isResumeBlocked()) {
+                definition.setDurabilityHint(FlowDurabilityHint.PERFORMANCE_OPTIMIZED);
```
jglick: This is not right. `FlowDefinition`s are `Describable`s. Thus they must never have state that is modified except in direct response to a user saving a configuration screen or the like. Anyway a single definition might in principle be shared among any number of jobs, not to mention builds.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -321,6 +324,32 @@ public void setQuietPeriod(Integer seconds) throws IOException {
         return getProperty(DisableConcurrentBuildsJobProperty.class) == null;
     }
 
+    @Exported
+    public boolean isResumeBlocked() {
+        return getProperty(DisableResumeJobProperty.class) != null;
+    }
+
+    public void setResumeBlocked(boolean resumeBlocked)  {
```
jglick: Can be deleted with a proper extension point, as discussed elsewhere.

src/main/java/org/jenkinsci/plugins/workflow/job/properties/DisableResumeJobProperty.java
svanoort: That's [JENKINS-47390](https://issues.jenkins-ci.org/browse/JENKINS-47390) -- a separate work item.

jglick:
merge conflicts

svanoort:
@jglick Merge conflicts are new from last night's work, and now resolved

src/main/resources/org/jenkinsci/plugins/workflow/job/properties/DurabilityHintJobProperty/config-details.jelly
```diff
@@ -23,13 +23,14 @@ THE SOFTWARE.
...
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
-    <f:entry title="${%Custom Durability Setting}" field="hint">
+    <f:entry title="${%Custom Pipeline Speed/Durability Level}" field="hint">
         <j:set var="defaultHint" value="${descriptor.defaultDurabilityHint}"/>
         <select class="setting-input" name="hint">
             <j:forEach var="it" items="${descriptor.durabilityHintValues}">
-                <f:option value="${it.name()}" selected="${instance != null ? it.name==instance[field].name() : it.name==defaultHint}">
-                    ${it.description}
-                </f:option>
+                <j:set var="optionBody" encode="false">${it.description}</j:set>
```
jglick: Skip this and inside `<option>` just write

```xml
<j:out value="${it.description}"/>
```

src/main/resources/org/jenkinsci/plugins/workflow/job/properties/DurabilityHintJobProperty/help.html
```diff
@@ -22,7 +22,29 @@
  ~ THE SOFTWARE.
  -->
 <div>
-    Durability hints that guide how we trade off between performance and durability against failures.
+    <p>
+        This setting allows users to change the default durability mode for running Pipelines.  In most cases this is a
+        trade-off between performance and the ability for running pipelines to resume after unplanned Jenkins outages.
 
-    <p>This may also be set globally under Manage Jenkins > Configure System.
+        <br><br><strong>What does this do?</strong>
```
jglick: Does this not duplicate help text in `workflow-api`? That is not going to be maintainable.

If you need to write up an elaborate explanation, it is perhaps better to define something like https://jenkins.io/redirect/pipeline-durability/ which could be a wiki page, article, whatever (can be changed at any time), and let each point in the UI which touches on this link to it for all details.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -218,7 +218,7 @@ private void assertColor(WorkflowRun b, BallColor color) throws IOException {
         b1.delete();
         b1 = null;
         r.jenkins.getQueue().clearLeftItems(); // so we do not need to wait 5m
-        MemoryAssert.assertGC(b1r, false);
+//        MemoryAssert.assertGC(b1r, false);
```
jglick: Uh…

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -259,9 +266,22 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
...
             }
+
             Owner owner = new Owner(this);
-            
             FlowExecution newExecution = definition.create(owner, listener, getAllActions());
+
+            boolean loggedHintOverride = false;
+            if (getParent().isResumeBlocked()) {
+                if (newExecution instanceof BlockableResume) {
+                    ((BlockableResume) newExecution).setResumeBlocked(true);
```
jglick: This looks unnecessary. The `FlowExecution` should be able to determine for itself whether or not resume is blocked, by inspecting the durability hint on its owner. That is why I say that you probably do not need a `setResumeBlocked` method.

svanoort: @jglick This is a separate feature -- https://issues.jenkins-ci.org/browse/JENKINS-33761 -- you can still resume a build with the lowest durability hint as long as the master shut down properly.

jglick:
From the beta announcement:

> use a "properties" step - the setting will apply to the NEXT run after the step is executed

If you accept my suggestion in this and other PRs to make the durability hint “pull” rather than “push” and deleting `BlockableResume`, then applying this setting using `properties` would take effect immediately in the running build. Since this is typically the first thing the build does, even in Declarative AFAIK, it is unlikely there is anything worth saving prior to this anyway.

That would also mean the branch property becomes much less relevant, since `Jenkinsfile` customization suffices, and you could perhaps simplify things by just throwing out https://github.com/jenkinsci/workflow-multibranch-plugin/pull/68. There is a longstanding “Pipeline-as-code” design policy of only allowing branch properties to be applicable to Pipeline when it is logically impossible for the feature to be applied during a running build and the feature must be supported on build 1 of a new branch project—which is why the only such property today (that I know of) is the one to suppress automatic triggering of builds on a newly created branch project.

`properties` could also be used to disable resumption in a running build. No “yield” should have occurred in the CPS VM thread prior to the completion of the `properties` step in a typical case, since it is synchronous blocking, and it is typically the first step run (or only preceded by `echo`es, also synch blocking); and a polite shutdown waits for all CPS VM threads to yield, which would allow all builds which have already started to set this job property (and the `config.xml` save would have been committed.) That leaves the unlikely case of a hard kill in the short interval after the build has started but before the `Job.save` completes. In that case, if any committed `program.dat` metadata claims the `properties` step has not yet run, then after a resume the step will run again and _subsequent_ restarts will kill the build as expected. This is OK, since presumably whatever problems caused you to disable resumption—bugs in `node`/`sh`, usually—would not yet have had an opportunity to be encountered. So could `program.dat` be written claiming the step has already run _before_ it does run? Seems impossible, because it would need to be recording the return value or throwable, which has not yet been produced.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: The same consideration applies to both features I think, as detailed in the long-form comment.

svanoort:
> then applying this setting using properties would take effect immediately in the running build

@jglick This actually poses a serious problem, which is why this is "push" rather than "pull" (and we have the BranchProperty, per policy of using it when settings can't be changed mid-build).  There's a couple reasons that mean the setting needs to be fixed rigidly when the build begins:

1. FlowNodeStorage - since we have different implementations, caching, and deferred writes changing this on the fly will be intrinsically unsafe and brittle.
2. Assumptions used in persistence calls - in the runtime implementation (WF-CPS) we'd have to check for changes to persistence state and then either flush to disk or figure out whether or not we're in a "safely" persisted state currently and modify that. 
3.  These settings are checked in a variety of places and potentially in different threads, so making them runtime-mutable opens up risks of synchronization and threading problems.

By making the durability setting static for a given Run & execution, we are able to implement this with higher confidence in its robustness and make some simplifying assumptions in the logic.  

Being able to (potentially) toggle resume on/off at runtime on the other hand *only* has impacts at resume time -- so it's much simpler to make this mutable and safe to make that "pull" rather than "push".  This also has significant value for users, who may decide that a project isn't safe to resume after seeing issues with one build.  It's also handy for an administrator because you could use a groovy script to mark pipeline projects to not resume at restart, to ensure a quieter startup.

svanoort:
Test  failure appears to be just a timeout in the build process (sigh).

svanoort:
Test failure appears to tie to the InsaneEngine (memory analysis) taking too long + timeout of the build agent (maybe slow repo issues again).  I think safe to probably disregard.

jglick:
Looks like you may have introduced a memory leak here? Needs to be investigated.

svanoort:
@jglick How do you get that from this?  What I'm seeing is similar to the failures sometimes seen with IDE runs in analyzing objects in memory with that Insane library (their name, not mine). 

> [ERROR] cleanup(org.jenkinsci.plugins.workflow.job.WorkflowRunTest)  Time elapsed: 180.064 s  <<< ERROR! 
        org.junit.runners.model.TestTimedOutException: test timed out after 180 seconds
	at org.netbeans.insane.impl.LiveEngine.visitRef(LiveEngine.java:123)
        at org.netbeans.insane.impl.LiveEngine.visitObjectReference(LiveEngine.java:113)
	at org.netbeans.insane.impl.InsaneEngine.processObject(InsaneEngine.java:295)
	at org.netbeans.insane.impl.InsaneEngine.process(InsaneEngine.java:201)
	at org.netbeans.insane.impl.InsaneEngine.traverse(InsaneEngine.java:97)
	at org.netbeans.insane.impl.LiveEngine.traceImpl(LiveEngine.java:188)
	at org.netbeans.insane.impl.LiveEngine.trace(LiveEngine.java:166)
	at org.jvnet.hudson.test.MemoryAssert.fromRoots(MemoryAssert.java:292)
	at org.jvnet.hudson.test.MemoryAssert.assertGC(MemoryAssert.java:178)
	at org.jenkinsci.plugins.workflow.job.WorkflowRunTest.cleanup(WorkflowRunTest.java:221)

Doesn't actually show a memory leak, just that the test itself is taking forever (not too surprising when you are doing deep memory analysis each time).

jglick:
Normally the reference gets cleared after the first full GC cycle (1330 IIRC), so the test completes fairly quickly. Here it is not getting cleared, which typically means you have introduced a memory leak; the test times out because it is trying more and more strenuously to force a particular object to be collected.

svanoort:
@jglick If there's an issue, it's in workflow-cps -- that's why I haven't marked it for review, because I'm verifying the memory behavior there.  It really feels like these memory tests could be more helpful in terms of making it possible to diagnose issues though.

jglick:
> I'm verifying the memory behavior there.

Yeah, that is what I suspected.

> these memory tests could be more helpful

Believe me, I have tried.

Normally a failure will be accompanied by reference chains after each collection showing what your leak is. I have not seen that here, which could mean a bug in the tool, or it could mean that a root reference is coming from something opaque to the tool, such as a local variable held in some thread's stack (there is just no way to introspect these that I know of).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -996,7 +996,10 @@ public int hashCode() {
                 copyLogs();
                 logsToCopy.put(node.getId(), 0L);
             }
-            node.addAction(new TimingAction());
+
+            if (node.getPersistentAction(TimingAction.class) == null) {
+                node.addAction(new TimingAction());
```
jglick: Possible race condition—might need to be inside the `completed` lock.

jglick:
> FlowNodeStorage - since we have different implementations, caching, and deferred writes changing this on the fly will be intrinsically unsafe and brittle.

I think I missed before that you have `BulkFlowNodeStorage` with a different file layout depending on the durability hint.

I am not sure if there are other aspects of durability which could be changed on the fly in a build like resumption can be (i.e., where each decision is logically isolated from the others), but for purposes of flow node storage it would be fine to just say that the storage engine is pinned when the build starts, according to job properties then in effect. Thus for a branch project using `properties`, build 1 would use some system default storage, but subsequent builds would use the selection.

This is akin to the treatment of parameter definitions, and in Declarative you have the opportunity of applying the selection during branch project creation so it applies even to build 1; see [JENKINS-41929](https://issues.jenkins-ci.org/browse/JENKINS-41929). That would avoid the need for a `DurabilityHintBranchProperty`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: I don't *think* it's possible to invoke twice, because in workflow-cps we add the action before GraphListeners get a chance to fire on it. 

The plus side it is that adding the action twice only adds a tiny bit of storage waste -- we look for the first action matching the class type.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -209,6 +210,7 @@ private void assertColor(WorkflowRun b, BallColor color) throws IOException {
         assertSame(color, b.getParent().getIconColor());
     }
 
+    @Ignore(value = "The framework used for memory testing here seems to be functionally broken and all signs from other tests are that we do NOT have memory leaks.")
```
jglick: I am not comfortable ignoring a test which solved a demonstrable memory leak merely because you have not yet figured out why your changes cause it to fail.

jglick: > why your changes cause it to fail

Not sure of this either, but it seems to involve running this test within the whole suite—it passes as a single test case.

I think you want to apply the same workaround as in https://github.com/jenkinsci/workflow-cps-plugin/pull/178: move this test to a separate suite, then configure Surefire to create a fresh JVM for each suite.

jglick: FTR this test was added in #44.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -65,6 +65,7 @@
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
 import static org.junit.Assert.*;
 import org.junit.ClassRule;
+import org.junit.Ignore;
```
jglick: unused import

PR #84, opened 2018-01-21
Clear out the logPrefixCache at end of build

svanoort:
Optimization noted when doing memory use profiling with Durability settings - we carry around this cache that we no longer need when the build is done. 

@reviewbybees


svanoort:
Supplanted by https://github.com/jenkinsci/workflow-job-plugin/pull/91

PR #83, opened 2017-12-21
Revert "Merge pull request #82" - restores cleanup code that requires core 2.76

svanoort:
This re-applies the code cleanup from PR  #67 which was reverted in #82 

This PR is so we don't forget to apply that cleanup -- but should be blocked until LTS 2.89.x has wider uptake (give it a month or two).

PR #85, opened 2018-01-22
[JENKINS-46652] Integration test

jglick:
Downstream of https://github.com/jenkinsci/jenkins/pull/3254.

@reviewbybees


jglick:
You can use the `do-not-merge` label for such cases. (`on-hold` would be better though we lack such a label in this repo.)

jglick:
If merging, continue with #83.

PR #87, opened 2018-01-26
Fix tests broken by [JENKINS-47393] when the CommandLauncher was removed from core

svanoort:
Found in testing for any net-new JEP-200 complaints in the tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -191,6 +191,12 @@
             <version>1.6</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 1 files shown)


svanoort:
@reviewbybees

dwnusbaum:
This can be closed, right? The current baseline is 2.121, and we don't have a command launcher dependency, but the tests are passing. I think the fix we needed was in [jenkins-test-harness 2.31](https://github.com/jenkinsci/jenkins-test-harness/#231-2017-oct-17), which is automatically picked by the parent pom in [version 2.37+](https://github.com/jenkinsci/plugin-pom/blob/master/CHANGELOG.md#237).

PR #88, opened 2018-02-08
Support environment variables set in Node Properties

atikhono:
Part of JENKINS-43894


atikhono:
@reviewbybees

atikhono:
This patch only works on master when retrieving Jenkinsfile from scm. In other cases there's no node context in WorkflowRun.getEnvironment, therefore no node properties. I will get back with another patch.

atikhono:
@jglick How can I access current node in `WorkflowRun.getEnvironment`, is that possible at all? I was told that `Computer.currentComputer` does not work for non-Freestyle projects.

atikhono:
Ping @jenkinsci/code-reviewers @jglick @abayer

atikhono:
https://issues.jenkins-ci.org/browse/JENKINS-36267

> The environment applies to the whole build. It makes no sense to return node-specific environment here. There might be multiple node blocks running in parallel!

PR #89, opened 2018-02-09
[JEP-206] Always use UTF-8 for the main Pipeline log file

jglick:
[JEP-206](https://github.com/jenkinsci/jep/blob/master/jep/206/README.adoc)

Subsumes #102. Downstream of https://github.com/jenkinsci/workflow-support-plugin/pull/56. A tiny piece pulled out of #27 as _one component of_ a solution to [JENKINS-31096](https://issues.jenkins-ci.org/browse/JENKINS-31096). Without https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/64, you will get mojibake, but possibly no worse than you already do. Compare the attempt to do something similar for freestyle in https://github.com/jenkinsci/jenkins/pull/3231.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
         <jenkins.version>2.73.3</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
-        <workflow-support-plugin.version>2.19-rc287.918f5958551d</workflow-support-plugin.version> <!-- TODO https://github.com/jenkinsci/workflow-support-plugin/pull/56 -->
+        <workflow-support-plugin.version>2.20</workflow-support-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
         <git-plugin.version>3.2.0</git-plugin.version>
         <jenkins-test-harness.version>2.33</jenkins-test-harness.version>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.36</version>
+        <version>3.4</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -62,11 +62,11 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>2.62</jenkins.version>
+        <jenkins.version>2.73.3</jenkins.version>
```
jglick: Unrelated, but easier to remember LTS baselines.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -465,4 +465,12 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
         }
     }
 
+    @Issue("JENKINS-31096")
+    @Test public void unicode() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        String message = "¡Čau → there!";
+        p.setDefinition(new CpsFlowDefinition("echo '" + message + "'", true));
+        r.assertLogContains(message, r.buildAndAssertSuccess(p));
```
jglick: On Windows (ISO-8859-1), this fails with mojibake with the `charset = "UTF-8"` line commented out, or with the upstream PR not integrated.

pom.xml
svanoort: Let's not bump the baseline unless there's a need to - this limits who can take advantage of new plugin releases.

pom.xml
```diff
@@ -62,11 +62,11 @@
...
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
-        <workflow-support-plugin.version>2.17</workflow-support-plugin.version>
-        <scm-api-plugin.version>2.1.1</scm-api-plugin.version>
+        <workflow-support-plugin.version>2.19-20180209.204154-1</workflow-support-plugin.version> <!-- TODO https://github.com/jenkinsci/workflow-support-plugin/pull/56 -->
+        <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
```
svanoort: Why?

svanoort:
"Without jenkinsci/workflow-durable-task-step-plugin#64, you will get mojibake, but possibly no worse than you already do." -- This worries me however.

pom.xml
jglick: Bumped only to the nearest LTS line. There is no reason for anyone to still be using intermediate weeklies. If they are, they need to update.

pom.xml
jglick: Something in the transitive deps needs it.

jglick:
IOW: if you are silly and only update certain plugins, things will not work very well. OTOH things are already broken, as can be seen from the bug report. If you just update all, things will be fine.

jglick:
> regressions on not-UTF systems where encodings were somehow aligned between components

Again, only if you selectively update plugins, pending [JENKINS-49651](https://issues.jenkins-ci.org/browse/JENKINS-49651).

PR #90, opened 2018-02-21
Add the CSS class "pipeline-annotated" to all pipeline annotated cons…

marcmac:
[JENKINS-41845] This will allow the administrator to override the display of these messages using the Simple Theme Plugin.


jglick:
Note that this is dead code as of #27 anyway.

PR #91, opened 2018-02-27
[JENKINS-38383] Avoid Stackoverflow and full-flowgraph scan when trying to get branch names for log prefixes

svanoort:
Solves [JENKINS-38383](https://issues.jenkins-ci.org/browse/JENKINS-38383) which could trigger errors upon restart with a complex Pipeline. 

Renames the `getPrefixCache` & its cache to better reflect what they do (get the branch name for a `FlowNode`). 

Also subsumes https://github.com/jenkinsci/workflow-job-plugin/pull/84 in passing since it's trivial and would cause a merge conflict there anyway. 

Should incidentally fix [this issue noted in JENKINS-50350](https://issues.jenkins-ci.org/browse/JENKINS-50350?focusedCommentId=332985&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-332985).

I *think* this will also avoid some really grotesque failures seen by organizations as a result of quirks in `getLogPrefix` & the error handling in the previous `LoadingCache`.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -25,9 +25,8 @@
 package org.jenkinsci.plugins.workflow.job;
 
 import com.google.common.base.Optional;
+import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -120,10 +119,10 @@ import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.flow.GraphListener;
 import org.jenkinsci.plugins.workflow.flow.StashManager;
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
+import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
 import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.console.WorkflowConsoleLogger;
-import org.jenkinsci.plugins.workflow.job.properties.DurabilityHintJobProperty;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
@@ -504,7 +503,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                     long old = entry.getValue();
                     OutputStream logger;
 
-                    String prefix = getLogPrefix(node);
+                    String prefix = getBranchName(node);
                     if (prefix != null) {
                         logger = new LogLinePrefixOutputFilter(listener.getLogger(), "[" + prefix + "] ");
                     } else {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -559,33 +558,61 @@ private long writeRawLogTo(AnnotatedLargeText<?> text, long start, OutputStream
     }
 
     @GuardedBy("completed")
-    private transient LoadingCache<FlowNode,Optional<String>> logPrefixCache;
-    private @CheckForNull String getLogPrefix(FlowNode node) {
-        // TODO could also use FlowScanningUtils.fetchEnclosingBlocks(node).filter(FlowScanningUtils.hasActionPredicate(ThreadNameAction.class)),
-        // but this would not let us cache intermediate results
+    private transient Cache<FlowNode,Optional<String>> branchNameCache;  // TODO Consider making this a top-level FlowNode API
```
svanoort: See below for why we can't use a LoadingCache anymore.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -559,33 +558,61 @@ private long writeRawLogTo(AnnotatedLargeText<?> text, long start, OutputStream
...
+
+    private Cache<FlowNode, Optional<String>> getBranchNameCache() {
```
svanoort: More for clealiness than anything else.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -559,33 +558,61 @@ private long writeRawLogTo(AnnotatedLargeText<?> text, long start, OutputStream
...
+        if (output != null) {
+            return output.orNull();
+        }
+
+        // We must explicitly check for the current node being the start/end of a parallel branch
+        if (node instanceof BlockEndNode) {
+            output = Optional.fromNullable(getBranchName(((BlockEndNode) node).getStartNode()));
+            cache.put(node, output);
+            return output.orNull();
+        } else if (node instanceof BlockStartNode) { // And of course this node might be the start of a parallel branch
```
svanoort: This trivial conditional check should massively reduce the getPersistentAction calls and improve performance a bit.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -559,33 +558,61 @@ private long writeRawLogTo(AnnotatedLargeText<?> text, long start, OutputStream
...
+                String name = threadNameAction.getThreadName();
+                cache.put(node, Optional.of(name));
+                return name;
+            }
+        }
+
+        // Check parent which will USUALLY result in a cache hit, but improve performance and avoid a stack overflow by not doing recursion
+        List<FlowNode> parents = node.getParents();
+        if (!parents.isEmpty()) {
+            FlowNode parent = parents.get(0);
```
svanoort: In practice we will never hit a case where parents > 1, because only the closing block for a Parallel will have > 1 parents, and because that's a BlockEndNode so it would be handled above.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -559,33 +558,61 @@ private long writeRawLogTo(AnnotatedLargeText<?> text, long start, OutputStream
...
+            output = cache.getIfPresent(parent);
+            if (output != null) {
+                cache.put(node, output);
+                return output.orNull();
+            }
+        }
+
+        // Fall back to looking for an enclosing parallel branch... but using more efficient APIs and avoiding a stack overflows
+        output = Optional.absent();
+        for (BlockStartNode myNode : node.iterateEnclosingBlocks()) {
```
svanoort: This is the meat of the fix.  I'll call it out here in case someone else spots it that we're not consulting the cache here, but that's okay because the cache can't help us anyway -- this logic only runs when we're starting or resuming a build (otherwise we'd have a parent cached).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -559,33 +558,61 @@ private long writeRawLogTo(AnnotatedLargeText<?> text, long start, OutputStream
...
+                cache.put(node, Optional.of(name));
+                return name;
+            }
+        }
+
+        // Check parent which will USUALLY result in a cache hit, but improve performance and avoid a stack overflow by not doing recursion
+        List<FlowNode> parents = node.getParents();
+        if (!parents.isEmpty()) {
+            FlowNode parent = parents.get(0);
+            output = cache.getIfPresent(parent);
```
svanoort: This one line is why we can't use a `LoadingCache` anymore -- we can't reference the `Cache` itself from the `CacheLoader` implementation.  So we need to explicitly work with the cache directly.

svanoort: This line is also 100% necessary though, because it lets us still get the full benefits of the prior caching approach, while not triggering unbounded recursive calls if trying to get a prefix for the parent.

svanoort:
@reviewbybees

svanoort:
@jglick Pick one of these two:  

1. Either we never needed this feature (showing which branch of a parallel the log originates from) and I can just go ahead and remove it entirely with great satisfaction **OR**
2. It's something that #27 proposes to remove which is shouldn't be -- one of many reasons why it will need some significant work to get it to a merge-ready state.

Either way it's a pretty painful problem *now* for people and wasn't hard to fix, so is the code change solid or not?

jglick:
> It's something that #27 proposes to remove

That. This whole code area is deleted and replaced by something very different.

> is the code change solid

No idea, I did not have time to look at it in any detail.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
rsandell: Why aren't these helpful comments in the code?

svanoort: @rsandell  A fair point -- the code comment about avoiding a stack overflow by not using recursion covers why it works the way it does (and should make anybody stop and think before switching to a LoadingCache again).  

The comment is in the PR itself because I'm explaining WHY the LoadingCache is going away -- so it's really only applicable to the code delta itself.

oleg-nenashev:
@svanoort @abayer any release ETA?

svanoort:
@oleg-nenashev "Soon"?  This *ought* to be run through a scalability test to ensure it doesn't regress performance too significantly but that might take a while.

oleg-nenashev:
@svanoort Any updates regarding "soon"? :)

svanoort:
Basic scalability testing comes up clean for me (with ParallelPlusSteps as the sample job).  It's not a comprehensive test it's enough to smoke-test that this doesn't introduce major regressions.

svanoort:
@jglick @oleg-nenashev  I'm just requesting a re-review since there were some merge conflicts introduced in a moderately sensitive area, but if you guys think that looks OK we're ready to merge here.

svanoort:
(for those who were wondering, the holdup was setting up an running enough of a scalability test + merge conflict resolution -- needed enough time to get decent data out of the test).

PR #93, opened 2018-03-28
Fix resume issues [JENKINS-49686] and [JENKINS-50199] and [JENKINS-50407]

svanoort:
Aims to resolve (in conjunction with workflow-cps changes as well that help harden the logic): 

* [JENKINS-50199](https://issues.jenkins-ci.org/browse/JENKINS-50199)
* [JENKINS-49686](https://issues.jenkins-ci.org/browse/JENKINS-49686)
* [JENKINS-50407](https://issues.jenkins-ci.org/browse/JENKINS-50407)

Includes lazy-load for FlowExecutions (enhancement):  [JENKINS-45585](https://issues.jenkins-ci.org/browse/JENKINS-45585) -- which if merged will replace [49086](https://issues.jenkins-ci.org/browse/JENKINS-49086) because it will defer a ton of the load-time logic for completed Pipelines.

TODOs:

* Cross-test with workflow-cps durability tests
* Fix issue with unitialized listener field for unable-to-resume flows (may mandate changes in wf-cps too)
* Hardening: consider forcing atomic write of XML file when re-saving with the completed field persisted -- even when normally non-atomic (guarantees we won't mangle original build record if an error occurs upon save).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -1022,10 +1022,12 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
             if (node instanceof FlowEndNode) {
                 finish(((FlowEndNode) node).getResult(), execution != null ? execution.getCauseOfFailure() : null);
             } else {
-                try {
-                    save();
-                } catch (IOException x) {
-                    LOGGER.log(Level.WARNING, null, x);
+                if (execution != null && execution.getDurabilityHint().isPersistWithEveryStep()) {
+                    try {
+                        save();
+                    } catch (IOException x) {
+                        LOGGER.log(Level.WARNING, null, x);
+                    }
                 }
             }
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -656,7 +656,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                 boolean completedStateNotPersisted = completed == null;
                 super.onLoad();
 
-                // TODO See if we can simply this, especially around interactions with 'completed'.
+                // TODO See if we can simplify this, especially around interactions with 'completed'.
 
                 if (execution != null && (completed == null || !completed.get())) {
                     FlowExecution fetchedExecution = getExecution();  // Triggers execution.onLoad so we can resume running if not done
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -89,7 +89,6 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -139,7 +138,7 @@ import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.interceptor.RequirePOST;
 
 @SuppressWarnings("SynchronizeOnNonFinalField")
-@SuppressFBWarnings(value="JLM_JSR166_UTILCONCURRENT_MONITORENTER", justification="completed is an unusual usage")
+@SuppressFBWarnings(value="RC_REF_COMPARISON_BAD_PRACTICE_BOOLEAN", justification="We need to appropriately handle null completion states from legacy builds.")
 public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements FlowExecutionOwner.Executable, LazyBuildMixIn.LazyLoadingRun<WorkflowJob,WorkflowRun>, RunWithSCM<WorkflowJob,WorkflowRun> {
 
     private static final Logger LOGGER = Logger.getLogger(WorkflowRun.class.getName());
@@ -171,7 +170,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
 
     private transient boolean allowKill;
 
-    /** Controls whether or not our execution has been initialized via its {@link FlowExecution#onLoad(FlowExecutionOwner)} method yet if.*/
+    /** Controls whether or not our execution has been initialized via its {@link FlowExecution#onLoad(FlowExecutionOwner)} method yet.*/
     transient boolean executionLoaded = false;  // NonPrivate for tests
 
     /**
@@ -191,7 +190,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
      * Flag for whether or not the build has completed somehow.
      * This was previously a transient field, so we may need to recompute in {@link #onLoad} based on {@link FlowExecution#isComplete}.
      */
-    AtomicBoolean completed;  // Non-private for testing
+    Boolean completed;  // Non-private for testing
 
     private transient Object logCopyGuard = new Object();
 
@@ -308,7 +307,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
             synchronized (getLogCopyGuard()) {  // Technically safe but it makes FindBugs happy
                 FlowExecutionList.get().register(owner);
                 newExecution.addListener(new GraphL());
-                completed = new AtomicBoolean();
+                completed = Boolean.FALSE;
                 logsToCopy = new ConcurrentSkipListMap<>();
                 executionLoaded = true;
                 execution = newExecution;
@@ -382,7 +381,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                         // Loading run, give it a moment.
                         return;
                     }
-                    if (completed.get()) {
+                    if (completed) {
                         asynchronousExecution.completed(null);
                         copyLogsTask.get().cancel(false);
                         return;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -171,6 +171,9 @@ public String url() {
 
     private transient boolean allowKill;
 
+    /** Controls whether or not our execution has been initialized via its {@link FlowExecution#onLoad(FlowExecutionOwner)} method yet if.*/
+    transient boolean executionLoaded = false;  // NonPrivate for tests
```
abayer: I actually wonder if that might be something we want to expose outside of just tests - I don't have a specific scenario in mind, but I know I've stumbled across cases where this could be useful information. I think. I may be getting confused with something else, though.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -186,17 +189,14 @@ public String url() {
...
+     * This was previously a transient field, so we may need to recompute in {@link #onLoad} based on {@link FlowExecution#isComplete}.
      */
-    private transient AtomicBoolean completed;
+    AtomicBoolean completed;  // Non-private for testing
+
+    private transient Object logCopyGuard = new Object();
 
     /** map from node IDs to log positions from which we should copy text */
-    private Map<String,Long> logsToCopy;
+    Map<String,Long> logsToCopy;
```
abayer: Probably worth a `// Non-private for testing` comment, but not strictly necessary since that's basically the only reason anything we do ever is package-private.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -206,6 +206,30 @@ public String url() {
...
     private transient boolean firstTime;
 
+    private synchronized Object getLogCopyGuard() {
+        if (logCopyGuard == null) {
+            logCopyGuard = new Object();
+        }
+        return logCopyGuard;
+    }
+
+    /** Avoids creating new instances, analogous to {@link TaskListener} but as full StreamBuildListener. */
```
abayer: Think you meant `TaskListener.NULL` here?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Can always expose it with a getter.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -506,9 +535,9 @@ private void copyLogs() {
 
                     String prefix = getLogPrefix(node);
                     if (prefix != null) {
-                        logger = new LogLinePrefixOutputFilter(listener.getLogger(), "[" + prefix + "] ");
+                        logger = new LogLinePrefixOutputFilter(getListener().getLogger(), "[" + prefix + "] ");
```
jglick: Tons of merge conflicts with #27 coming up, sigh. Was the introduction of `getListener()` necessary to solve one of these bugs?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -679,49 +716,57 @@ private String key() {
...
+
         try {
-            save();
-        } catch (Exception x) {
-            LOGGER.log(Level.WARNING, "failed to save " + this, x);
-        }
-        Timer.get().submit(() -> {
+            LOGGER.log(Level.INFO, "{0} completed: {1}", new Object[]{toString(), getResult()});
+            if (listener == null) {
+                // Can we actually rely on this, now that we use an initializing getter?
```
jglick: Is this a TODO?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -622,50 +651,58 @@ private String key() {
...
+                        if (completed == null) {
+                            completed = new AtomicBoolean(fetchedExecution.isComplete());
+                        }
+
+                        if (!completed.get()) {
+                            // we've been restarted while we were running. let's get the execution going again.
+                            FlowExecutionListener.fireResumed(fetchedExecution);
+
+                            getListener().getLogger().println("Resuming build at " + new Date() + " after Jenkins restart");
+                            Timer.get().submit(new Runnable() { // JENKINS-31614
```
jglick: As long as you are shuffling code around, might as well:

```java
Timer.get().submit(() -> Queue.getInstance().schedule(new AfterRestartTask(this), 0)); // JENKINS-31614
```

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -622,50 +651,58 @@ private String key() {
...
+                if (completedStateNotPersisted && completed.get()) {
+                    try {
+                        save();
+                    } catch (Exception ex) {
+                        LOGGER.log(Level.WARNING, "Error while saving build to update completed flag", ex);
                     }
-                });
+                }
+            }
+        } finally {  // Ensure the run is ALWAYS removed from loading even if something failed, so threads awaken.
```
jglick: /me runs away in fear

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
```diff
@@ -93,8 +97,45 @@
...
+            assertNotNull(fe.getOwner());
+            assertNull(b.logsToCopy);
+        });
+    }
+
+    @Issue("JENKINS-45585")  // Verifies execution lazy-load
+    @Test public void lazyLoadExecution() {
+        story.thenWithHardShutdown(r -> {
+            WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("echo 'dosomething'", true));
```
jglick: Do you need to set the durability mode too? Best to not rely on defaults if it is significant here.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Something like that, more of a "I wonder what @jglick will say in review"

svanoort:
@jglick Does this help with Whitespace?  https://github.com/jenkinsci/workflow-job-plugin/pull/93/files?w=1

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -679,49 +708,52 @@ private String key() {
...
         }
-        logsToCopy = null;
+
         try {
-            save();
-        } catch (Exception x) {
-            LOGGER.log(Level.WARNING, "failed to save " + this, x);
-        }
-        Timer.get().submit(() -> {
+            LOGGER.log(Level.INFO, "{0} completed: {1}", new Object[]{toString(), getResult()});
```
rsandell: Is `Level.INFO` really wise here?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -679,49 +708,52 @@ private String key() {
...
+            if (listener == null) {
```
rsandell: what about `getListener()`?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -152,8 +151,8 @@ public String url() {
         }
     }
 
-    /** null until started, or after serious failures or hard kill */
-    private @CheckForNull FlowExecution execution;
+    /** Null until started, or after serious failures or hard kill. */
+    @CheckForNull FlowExecution execution; // Not private for test use only
```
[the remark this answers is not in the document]
svanoort: Nah, it's package-private, so it's already more restricted than if I used `@Restricted` on it

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -206,6 +205,30 @@ public String url() {
...
+            logCopyGuard = new Object();
+        }
+        return logCopyGuard;
+    }
+
+    /** Avoids creating new instances, analogous to {@link TaskListener#NULL} but as full StreamBuildListener. */
+    static final StreamBuildListener NULL_LISTENER = new StreamBuildListener(new NullStream());
+
+    /** Used internally to ensure listener has been initialized correctly. */
+    StreamBuildListener getListener() {
```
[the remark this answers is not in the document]
svanoort: Good catch. I believe there was a specific reason why it wasn't based on testing but... darned if I can remember the issue/exact chain of reasoning since it was rather complex.  

Give me a few, I'll either synchronize or provide a detailed comment to explain the rationale why not to.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Similar to the comment on synchronizing `getListener()` -- give me a bit and I'll either have it changed or better documented in code comments.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Well, that's what it was at before -- it should only run once per build (or at most twice in rare circumstances if something fails catastrophically right in the middle of the finalization).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -658,38 +661,31 @@ private String key() {
...
+                if (execution != null && completed != Boolean.TRUE) {
                     FlowExecution fetchedExecution = getExecution();  // Triggers execution.onLoad so we can resume running if not done
 
                     if (fetchedExecution != null) {
                         fetchedExecution.addListener(new GraphL());
                         executionPromise.set(fetchedExecution);
 
                         if (completed == null) {
-                            completed = new AtomicBoolean(fetchedExecution.isComplete());
+                            completed = Boolean.valueOf(fetchedExecution.isComplete());
```
jglick: Note: can just drop the `valueOf` call as this is done by autoboxing anyway.

svanoort:
Merging now that I've done one more round of integration testing locally using the related [workflow-cps PR](https://github.com/jenkinsci/workflow-cps-plugin/pull/216) and an updated snapshot.

PR #95, opened 2018-04-19
Protect against FlowExecutionOwner getting null execution due to lazy-load execution issues, also NPE-protect the executionPromise

svanoort:
Part of [JENKINS-50784](https://issues.jenkins-ci.org/browse/JENKINS-50784).

While testing the ReplayAction I discovered we were hitting cases where the execution couldn't be loaded by via `FlowExecutionOwner.getOrNull()` **after** lazy load was triggered.  This is the solution, with a testcase.

Plus it ensures the programPromise is null-safe and, um, actually has values set for the promise.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -157,6 +157,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     /**
      * {@link Future} that yields {@link #execution}, when it is fully configured and ready to be exposed.
      */
+    @CheckForNull
     private transient SettableFuture<FlowExecution> executionPromise = SettableFuture.create();
 
     private transient final LazyBuildMixIn.RunMixIn<WorkflowJob,WorkflowRun> runMixIn = new LazyBuildMixIn.RunMixIn<WorkflowJob,WorkflowRun>() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -315,16 +316,21 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
                 executionLoaded = true;
                 execution = newExecution;
             }
-
-            newExecution.start();
-            executionPromise.set(newExecution);
-            FlowExecutionListener.fireRunning(execution);
+            SettableFuture<FlowExecution> exec = getSettableExecutionPromise();
+            if (!exec.isDone()) {
+                exec.set(newExecution);
```
jglick: Minor race condition I suppose, though presumably in `run` of a new build there could be no confusion about which value is correct.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -315,16 +316,21 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
...
+            }
+            newExecution.start();  // We should probably have the promise set before beginning, no?
```
jglick: Seems prudent.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -466,7 +472,7 @@ public void doKill() {
         execution = null; // ensures isInProgress returns false
         FlowInterruptedException suddenDeath = new FlowInterruptedException(Result.ABORTED);
         finish(Result.ABORTED, suddenDeath);
-        executionPromise.setException(suddenDeath);
+        getSettableExecutionPromise().setException(suddenDeath);
```
jglick: No `isDone` check here?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1037,9 +1060,11 @@ private String key() {
         }
         @Override public FlowExecution getOrNull() {
             try {
-                ListenableFuture<FlowExecution> promise = run().getExecutionPromise();
+                WorkflowRun run = run();
+                ListenableFuture<FlowExecution> promise = run.getExecutionPromise();
                 if (promise.isDone()) {
-                    return promise.get();
+                    // Weird, I know, but this ensures we trigger the onLoad for the execution via the lazy-load mechanism
```
jglick: So IIUC this would benefit historical callers or `getOrNull` which might try the call again shortly, such as GUI elements.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Good question -- per the JavaDocs: "This method will return {@code true} if the exception was
   * successfully set, or {@code false} if the future has already been set or
   * cancelled."

PR #97, opened 2018-05-22
Fix thread contention with WorkflowRun

svanoort:
Solves [JENKINS-51377](https://issues.jenkins-ci.org/browse/JENKINS-51377) mostly by ensuring the most common operations (things like getExecution) only block when they need to. 

Strategies:

- Leverage  [the Only Safe Form of double-checked locking](https://wiki.sei.cmu.edu/confluence/display/java/LCK10-J.+Use+a+correct+form+of+the+double-checked+locking+idiom) with `volatile` fields for some lazily-initialized `transient` fields rather than full synchronization - since we have almost exclusively *read* operations happening this greatly reduces thread contention.  The write operations there are protected by marking the fields volatile where simply nulling them or using synchronization for less trivial operations.

TODO:
* [X] Fix failures in the durability-fuzzing tests - AFAICT they're not regressions, just pre-existing.
* [x] Check if we need to synchronize on the `WorkflowRun` in the `#doTerm()` method
* [x] Check if we need to synchronize on the logCopyGuard in the `#save()` method to prevent mutation there (since we no longer synchronize on the getter for that guard object) -- solved by restoring synchronization on the getter.
* [x] Document order of locking for WorkflowRun vs logCopyGuard and verify order is enforced

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -450,7 +450,6 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         final Throwable x = new FlowInterruptedException(Result.ABORTED);
         FlowExecution exec = getExecution();
         if (exec == null) { // Already dead, just make sure statuses reflect that.
-            // FIXME do we need to synchronize on the WorkflowRun too?
             synchronized (getLogCopyGuard()) {
                 // Null execution means a hard-kill of the execution and build is by definition dead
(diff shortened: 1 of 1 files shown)


jmdh:
As we would like to benefit from this change, it would be great to get it into a released version. Does anyone have any ETA for that? Thanks!

svanoort:
@jmdh I've cut this as release 2.22 for workflow-job -- would be curious to see how your system plays after the changes.

jmdh:
Cool, thanks! I'll see if we can get this installed together with the new LTS next week.

jmdh:
We haven't seen any related issues so far as we can tell!

PR #98, opened 2018-06-07
Incrementalified

jglick:


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -285,7 +285,7 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
                 if (!auth.equals(Jenkins.ANONYMOUS)) {
                     name = ModelHyperlinkNote.encodeTo(User.get(name));
                 }
-                myListener.getLogger().println(hudson.model.Messages.Run_running_as_(name));
+                myListener.getLogger().println(/* hudson.model.Messages.Run_running_as_(name) */ "Running as " + name);
```
svanoort: Would really like the internationalization to be resolved somehow here -- not a blocker but nice to have.

jglick: Internationalization is up to translation contributors.

Anyway localizing messages printed to the build log has always been suspect and arguably Jenkins should stop doing it. There is no reason to expect that the locale of the server has anything to do with that of its users.

jglick:
> pending CI pass

It already passed.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: @jglick I understand internationalization is not a priority if you're an English speaker, but this is a pretty huge thing overseas.  This is the last PR I will approve that removes the built-in Messages support for internationalization after this one -- you need to find a solution if you want to do this.

That change needs to be included in future Incrementals PRs -- I consider that effort incomplete when it regresses I8N.

jglick: Again this has nothing whatsoever to do with Incrementals. It is just a side effect of updating a parent POM to something that actually enforces restrictions we had _intended_ to put in place long before but which were accidentally being ignored.

And again translations are handled by, well, translators. They routinely look for untranslated messages that they think _should_ be translated, and translate them. Not something English-speaking devs need to worry about other than merging translation PRs when they show up.

And again internationalization of things printed to the build log was probably wrong to begin with.

PR #102, opened 2018-07-19
[JENKINS-51869] Example of update to drop --first-parent

jglick:
Downstream of https://github.com/jenkinsci/plugin-pom/pull/119. Note that this has deployed `https://2.24-rc760.d0b56943b4cf` as opposed to `master`’s `2.24-rc353.b6a4049f4755`.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -2,6 +2,6 @@
   <extension>
     <groupId>io.jenkins.tools.incrementals</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.0-beta-3</version>
+    <version>1.0-beta-4</version>
   </extension>
 </extensions>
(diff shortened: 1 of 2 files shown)


jglick:
For JEP-206 + JEP-210 etc. this will in fact need to be applied to a number of other plugins as well: `workflow-api`, `workflow-support`, `workflow-durable-task-step`, & `durable-task` off the top of my head. So expect a bunch of parent update PRs.

svanoort:
@jglick Please provide some context when opening a PR - the description given may be meaningful to you but it means nothing to us, and the upstream isn't any more helpful on its own.

jglick:
See [JENKINS-51869](https://issues.jenkins-ci.org/browse/JENKINS-51869) for background.

jglick:
> Short, human-readable summary please?

The behavior of `git-changelist-maven-extension` prior to `1.0-beta-4` is undesirable when there are PRs “stacked” on top of others, as I noticed while developing JEP-210. This PR corrects that, by updating the extension.

If you are interested in technical details of Incrementals, please ask in the source PRs. From the perspective of plugin developers, this PR is as trivial as it looks: updating the parent POM.

PR #103, opened 2018-07-19
Delegate to nondeprecated StashManager overloads

jglick:
https://github.com/jenkinsci/workflow-api-plugin/pull/67 introduced some new overloads permitting a custom artifact manager to print messages when deleting artifacts or stashes, and deprecated the old ones, so this just switches to the new ones. For now we do not have a good place to send the messages so this PR merely allows them to be discovered via `FINE-` loggers, but at least it makes it clearer what code to edit to get access.


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -830,8 +830,8 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
         }
 
         try {
-            StashManager.maybeClearAll(this);
-        } catch (IOException x) {
+            StashManager.maybeClearAll(this, /* or move up before closing getListener()? */new LogTaskListener(LOGGER, Level.FINE));
```
jglick: Could have attempted to make such a change, but making a more conservative PR for now.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -842,7 +842,11 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
 
     @Override public void deleteArtifacts() throws IOException {
         super.deleteArtifacts();
-        StashManager.clearAll(this);
+        try {
+            StashManager.clearAll(this, /* core API defines no log sink for this */ new LogTaskListener(LOGGER, Level.FINE));
```
jglick: …since `BuildDiscarder` is called from a new build and then deletes artifacts from older builds. The `logRotate` call is currently also asynchronous and done after closing the logger for the new build in `finish`. Both things _could_ be changed, but we would still need a core API change to allow us to thread the right `TaskListener` through `Job.logRotate` → `BuildDiscarder.perform` (in `LogRotator`) → `Run.deleteArtifacts`. Also through `LogRotator.perform` → `Run.delete` → `Run.deleteArtifacts` as of https://github.com/jenkinsci/jenkins/pull/3484 (which is in 2.121.2). A lot of work just to display

> Deleted … stash(es) from …

Currently artifact deletion also not have any available `TaskListener`, though that is not touched by this PR.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -62,6 +62,7 @@
 import hudson.slaves.NodeProperty;
 import hudson.util.DaemonThreadFactory;
 import hudson.util.Iterators;
+import hudson.util.LogTaskListener;
```
jglick: Should be safely remotable since it delegates to a `StreamTaskListener` which in turn uses a `RemoteOutputStream` so that the original `LogOutputStream` need not be serialized. Should not matter for `JCloudsArtifactManager` anyway since its `clearAllStashes` does not need to remote the `TaskListener`: deletion happens from the master only.

jglick:
I do not see any `stable` branch here yet, but OTOH there has been plenty of stuff merged to `master` which is clearly not intended to be “stable”. :-)

PR #104, opened 2018-08-28
[JENKINS-50199] Fix cases where builds would "resume" even though the execution is complete

svanoort:
[JENKINS-50199](https://issues.jenkins-ci.org/browse/JENKINS-50199).

Should finally put the last variant of this problem to rest.  Root cause: execution completed (and this is saved), but something prevented saving the build *itself* with completed status.   Unfortunately we didn't do anything to detect this case and mark the build itself as completed if its execution was. 

Solving the save itself is a separate problem - since that can potentially be a result of I/O failures or abrupt shutdowns, we have to handle the possibility anyway.

Also adds in the suite of persistence problem tests from the CPS plugin, since *both* the CPS and workflow plugins need to have persistence logic correct internally (and we need to verify this since either can break it).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.25</version>
+            <version>2.27</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -92,19 +92,19 @@
             <!-- Satisfy upper bound dependencies -->
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.39</version>
+            <version>1.42</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.43</version>
+            <version>2.54</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.10</version>
+            <version>1.14</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -130,6 +130,12 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-durable-task-step</artifactId>
             <version>2.11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-input-step</artifactId>
+            <version>2.5</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -877,7 +877,12 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                         if (fetchedExecution.isComplete()) {  // See JENKINS-50199 for cases where the execution is marked complete but build is not
                             // Somehow arrived at one of those weird states
                             this.completed = true;
-                            setResult(Result.FAILURE);
+                            Result finalResult = Result.FAILURE;
+                            List<FlowNode> heads = fetchedExecution.getCurrentHeads();
+                            if (!heads.isEmpty() && heads.get(0) instanceof FlowEndNode) {
+                                finalResult = ((FlowEndNode)(heads.get(0))).getResult();
+                            }
+                            setResult(finalResult);
                             fetchedExecution.removeListener(finishListener);
                             saveWithoutFailing();
                         } else {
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -132,6 +132,12 @@
             <version>2.11</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-input-step</artifactId>
```
jglick: Use `SemaphoreStep` instead.

src/test/java/org/jenkinsci/plugins/workflow/job/CpsPersistenceTest.java
```diff
@@ -0,0 +1,432 @@
...
+        WorkflowJob job = j.jenkins.createProject(WorkflowJob.class, jobName);
+        job.setDefinition(new CpsFlowDefinition("input 'pause'", true));
+        job.addProperty(new DurabilityHintJobProperty(durabilityHint));
+
+        WorkflowRun run = job.scheduleBuild2(0).getStartCondition().get();
+        ListenableFuture<FlowExecution> listener = run.getExecutionPromise();
+        FlowExecution exec = listener.get();
+        while(exec.getCurrentHeads().isEmpty() || (exec.getCurrentHeads().get(0) instanceof FlowStartNode)) {  // Wait until input step starts
+            System.out.println("Waiting for input step to begin");
+            Thread.sleep(50);
```
jglick: More easily and reliably done with `SemaphoreStep`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -874,10 +874,24 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
                     }
                     fetchedExecution.onLoad(new Owner(this));
                     if (this.completed != Boolean.TRUE) {
-                        // Defer the normal listener to ensure onLoad can complete before finish() is called since that may
-                        // need the build to be loaded and can result in loading loops otherwise.
-                        fetchedExecution.removeListener(finishListener);
-                        fetchedExecution.addListener(new GraphL());
```
jglick: Note to self: minor merge conflict with https://github.com/jenkinsci/workflow-job-plugin/pull/27.

jglick:
Code change looks OK I suppose. As far as the tests go: if there is a particular test in `workflow-cps-plugin` which reproduces the stated problem if you were to update the `workflow-job` test dep, then simply revert all but the `src/main/java/` portion of this PR, wait for Incrementals deployment, and then bump the test dep in `workflow-cps-plugin`.

(And if there is no such test yet, then why are you copying all these tests here? Write a new test, here, to reproduce the problem, if you can.)

src/test/java/org/jenkinsci/plugins/workflow/job/CpsPersistenceTest.java
```diff
@@ -0,0 +1,432 @@
...
+            Field resultField = Run.class.getDeclaredField("result");
+            resultField.setAccessible(true);
+            resultField.set(run, null);
+
+            run.save();
+        });
+        story.then( j->{
+            WorkflowJob r = j.jenkins.getItemByFullName(DEFAULT_JOBNAME, WorkflowJob.class);
+            WorkflowRun run = r.getBuildByNumber(build[0]);
+            assertCompletedCleanly(run);
```
dwnusbaum: Maybe add `Assert.assertEquals(Result.SUCCESS, run.getResult());` to show that copying the result from the FlowEndNode works correctly?

src/test/java/org/jenkinsci/plugins/workflow/job/CpsPersistenceTest.java
```diff
@@ -0,0 +1,432 @@
...
+        });
+    }
+
+    @Test
+    @Issue("JENKINS-50199")
+    /** Replicates case where builds resume when the should not due to build's completion not being saved. */
+    public void completedExecutionButRunIncomplete() throws Exception {
```
svanoort: @jglick @dwnusbaum This is the net-new testcase.

svanoort:
@jglick We have long needed to have the tests in *both* places to catch regressions introduced in both Workflow-Job and Workflow-CPS persistence calls.  In the past we did indeed use a SNAPSHOT/incremental build of workflow-job and then did a bump to workflow-CPS -- but that's a slow and inefficient process where it's easy to slip in regressions by failing to run something against the latest build. 

Adapting existing tests from CPS (plus a new one) isn't perfect solution but when we need to be focused on velocity, refactoring tests should not be a high priority -- and we really do need that coverage.

jglick:
> when we need to be focused on velocity, refactoring tests should not be a high priority

I would rather argue that improving how/where tests are run is exactly what we need to do in order to increase velocity. Duplicating code is introducing technical debt which will be a drag on further work.

src/test/java/org/jenkinsci/plugins/workflow/job/CpsPersistenceTest.java
```diff
@@ -0,0 +1,432 @@
...
+        final int[] build = new int[1];
+        story.thenWithHardShutdown( j -> {
+            WorkflowRun run = runBasicPauseOnInput(j, DEFAULT_JOBNAME, build);
+            CpsFlowExecution cpsExec = (CpsFlowExecution)(run.getExecution());
+            InputStepExecution exec = getInputStepExecution(run, "pause");
+            exec.doProceedEmpty();
+            j.waitForCompletion(run);
```
jglick, line 384: see https://github.com/jenkinsci/jenkins-test-harness/pull/596

PR #109, opened 2018-10-06
Avoid deprecated api usage

darxriggs:
This is based on the analysis https://ci.jenkins.io/job/Infra/job/deprecated-usage-in-plugins/job/master/566/artifact/output/usage-by-plugin.html.


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -275,7 +275,10 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
             if (!auth.equals(ACL.SYSTEM)) {
                 String name = auth.getName();
                 if (!auth.equals(Jenkins.ANONYMOUS)) {
-                    name = ModelHyperlinkNote.encodeTo(User.get(name));
+                    User user = User.getById(name, false);
```
svanoort: This appears to be a separate fix... but a worthwhile one that avoids an NPE potentially.

src/main/java/org/jenkinsci/plugins/workflow/job/views/GraphVizAction.java
```diff
@@ -27,13 +27,13 @@
...
 import hudson.model.Action;
 import hudson.util.HttpResponses;
-import hudson.util.IOUtils;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.Collection;
 import java.util.Collections;
 import jenkins.model.TransientActionFactory;
+import org.apache.commons.io.IOUtils;
```
svanoort: I'm a little confused -- were we somehow importing the wrong dependency here?

PR #112, opened 2018-11-05
Optimized version of doConsoleText to avoid a memory buffer

jglick:
Noticed by @kshultzCB.


jglick:
hmm, CI issues

alok15ee:
I think this limitation is still not fixed. Parallel builds in the pipeline are still running into this issue.

jglick:
#164?

PR #113, opened 2018-11-05
Previous doConsoleText override failed to work past LargeText.MAX_LINES_READ

jglick:
Fix to #112 discovered by @kshultzCB.


jglick:
more CI problems

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1024,7 +1024,15 @@ public int hashCode() {
     @Override public void doConsoleText(StaplerRequest req, StaplerResponse rsp) throws IOException {
         rsp.setContentType("text/plain;charset=UTF-8");
         try (OutputStream os = rsp.getCompressedOutputStream(req)) {
-            getLogText().writeLogTo(0, os);
+            // Similar to writeWholeLogTo but terminates even if !logText.complete:
+            long pos = 0;
+            while (true) {
+                long pos2 = getLogText().writeLogTo(pos, os);
```
svanoort: Interesting possibility: what if the log is being generated faster than it can be written to output?

dwnusbaum:
`getLogInputStream` (a few lines up in the diff) has the same problem, right?

PR #114, opened 2018-11-05
Fix getLogInputStream and doConsoleText for large log files

dwnusbaum:
Subsumes #113. CC @kshultzCB 

I will add a test specifically for `getLogInputStream`.


dwnusbaum:
Happy to consolidate the fix/test if desired to avoid duplication.

dwnusbaum:
Just noticed that `getLog` has the same problem. I am not sure how often that method is called since it's deprecated, but I will go ahead and fix it as well.

PR #118, opened 2018-11-08
[JENKINS-54128] More efficient implementation of getLogFile that just uses the existing file when possible

jglick:
[JENKINS-54128](https://issues.jenkins-ci.org/browse/JENKINS-54128) Some plugins apparently call this a _lot_ so deal with it.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -109,6 +109,7 @@ import org.jenkinsci.plugins.workflow.flow.StashManager;
 import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.console.NewNodeConsoleNote;
+import org.jenkinsci.plugins.workflow.log.FileLogStorage;
 import org.jenkinsci.plugins.workflow.log.LogStorage;
 import org.jenkinsci.plugins.workflow.log.TaskListenerDecorator;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
@@ -1080,16 +1081,21 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
 
     @Override public File getLogFile() {
         LOGGER.log(Level.WARNING, "Avoid calling getLogFile on " + this, new UnsupportedOperationException());
-        try {
-            File f = File.createTempFile("deprecated", ".log", getRootDir());
-            f.deleteOnExit();
-            try (OutputStream os = new FileOutputStream(f)) {
-                getLogText().writeRawLogTo(0, os);
+        File f = super.getLogFile();
+        if (LogStorage.of(asFlowExecutionOwner()) instanceof FileLogStorage) {
+            // FileLogStorage does write a file with the same name and the same format as before JEP-210, so accept it if it is there.
+        } else {
+            // Some other storage. Cache under the traditional name, but load from the defined source on each call.
+            try {
+                f.deleteOnExit();
+                try (OutputStream os = new FileOutputStream(f)) {
+                    getLogText().writeRawLogTo(0, os);
+                }
+            } catch (IOException x) {
+                throw new RuntimeException(x);
             }
-            return f;
-        } catch (IOException x) {
-            throw new RuntimeException(x);
         }
+        return f;
     }
 
     static void alias() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -1089,7 +1089,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
             try {
                 f.deleteOnExit();
                 try (OutputStream os = new FileOutputStream(f)) {
-                    getLogText().writeRawLogTo(0, os);
+                    writeLogTo(getLogText()::writeRawLogTo, os);
                 }
             } catch (IOException x) {
                 throw new RuntimeException(x);
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -1087,12 +1087,11 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
             // FileLogStorage does write a file with the same name and the same format as before JEP-210, so accept it if it is there.
             // This is the normal case if no additional plugin is installed.
         } else {
-            // Some other storage, perhaps cloud-based. Cache under the traditional name, but load from the defined source on each call.
-            try {
-                f.deleteOnExit();
-                try (OutputStream os = new FileOutputStream(f)) {
-                    writeLogTo(getLogText()::writeRawLogTo, os);
-                }
+            // Some other storage, perhaps cloud-based. Load from the defined source on each call, but do not create more than one copy.
+            f = new File(f.getPath() + ".tmp");
+            f.deleteOnExit();
+            try (OutputStream os = new FileOutputStream(f)) {
+                writeLogTo(getLogText()::writeRawLogTo, os);
             } catch (IOException x) {
                 throw new RuntimeException(x);
             }
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -64,7 +64,6 @@ import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -109,7 +108,6 @@ import org.jenkinsci.plugins.workflow.flow.StashManager;
 import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.console.NewNodeConsoleNote;
-import org.jenkinsci.plugins.workflow.log.FileLogStorage;
 import org.jenkinsci.plugins.workflow.log.LogStorage;
 import org.jenkinsci.plugins.workflow.log.TaskListenerDecorator;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
@@ -1082,21 +1080,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     @Deprecated
     @Override public File getLogFile() {
         LOGGER.log(Level.WARNING, "Avoid calling getLogFile on " + this, new UnsupportedOperationException());
-        File f = super.getLogFile();
-        if (LogStorage.of(asFlowExecutionOwner()) instanceof FileLogStorage) {
-            // FileLogStorage does write a file with the same name and the same format as before JEP-210, so accept it if it is there.
-            // This is the normal case if no additional plugin is installed.
-        } else {
-            // Some other storage, perhaps cloud-based. Load from the defined source on each call, but do not create more than one copy.
-            f = new File(f.getPath() + ".tmp");
-            f.deleteOnExit();
-            try (OutputStream os = new FileOutputStream(f)) {
-                writeLogTo(getLogText()::writeRawLogTo, os);
-            } catch (IOException x) {
-                throw new RuntimeException(x);
-            }
-        }
-        return f;
+        return LogStorage.of(asFlowExecutionOwner()).getLogFile(this, !isLogUpdated());
     }
 
     static void alias() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -1020,7 +1020,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     @Override public InputStream getLogInputStream() throws IOException {
         // Inefficient but probably rarely used anyway.
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        writeLogTo(getLogText()::writeLogTo, baos);
+        writeLogTo(getLogText()::writeRawLogTo, baos);
         return new ByteArrayInputStream(baos.toByteArray());
     }
 
@@ -1085,8 +1085,9 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         File f = super.getLogFile();
         if (LogStorage.of(asFlowExecutionOwner()) instanceof FileLogStorage) {
             // FileLogStorage does write a file with the same name and the same format as before JEP-210, so accept it if it is there.
+            // This is the normal case if no additional plugin is installed.
         } else {
-            // Some other storage. Cache under the traditional name, but load from the defined source on each call.
+            // Some other storage, perhaps cloud-based. Cache under the traditional name, but load from the defined source on each call.
             try {
                 f.deleteOnExit();
                 try (OutputStream os = new FileOutputStream(f)) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1080,16 +1081,21 @@ private void writeLogTo(WriteMethod method, OutputStream os) throws IOException
...
-                getLogText().writeRawLogTo(0, os);
+        File f = super.getLogFile();
+        if (LogStorage.of(asFlowExecutionOwner()) instanceof FileLogStorage) {
+            // FileLogStorage does write a file with the same name and the same format as before JEP-210, so accept it if it is there.
+        } else {
+            // Some other storage. Cache under the traditional name, but load from the defined source on each call.
+            try {
+                f.deleteOnExit();
+                try (OutputStream os = new FileOutputStream(f)) {
+                    getLogText().writeRawLogTo(0, os);
```
dwnusbaum: Is this call also broken for long log files for in-progress builds? It seems like it would be most commonly used for complete builds, so probably doesn't matter, but you could use the following instead:

```suggestion
                    writeLogTo(getLogText()::writeRawLogTo, os);
```

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1080,16 +1081,21 @@ private void writeLogTo(WriteMethod method, OutputStream os) throws IOException
 
     @Override public File getLogFile() {
         LOGGER.log(Level.WARNING, "Avoid calling getLogFile on " + this, new UnsupportedOperationException());
```
dwnusbaum: Given the number of existing callers, should we only log the stack trace at FINE level?

jglick: Well…we only _know_ about those callers because it is at `WARNING`. And this would be important e.g. for getting sentry.io to show violations in Evergreen.

dwnusbaum:
Only other thing that came to mind while looking at this is that using the same file for non-FileLogStorage implementations seems like it could lead to strange bugs in the face of concurrent calls to `getLogFile` because it will be overwritten, but I couldn't really think of any better way to handle that.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1080,16 +1081,21 @@ private void writeLogTo(WriteMethod method, OutputStream os) throws IOException
...
-            f.deleteOnExit();
-            try (OutputStream os = new FileOutputStream(f)) {
-                getLogText().writeRawLogTo(0, os);
+        File f = super.getLogFile();
+        if (LogStorage.of(asFlowExecutionOwner()) instanceof FileLogStorage) {
+            // FileLogStorage does write a file with the same name and the same format as before JEP-210, so accept it if it is there.
+        } else {
+            // Some other storage. Cache under the traditional name, but load from the defined source on each call.
+            try {
+                f.deleteOnExit();
```
svanoort: Hey, isn't `f` potentially the original log file, not a deprecated temp file that we're copying to/from?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Isn't this potentially going to generate a log of logspam though?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Which means we'd... delete the build log on exit?  Unless there's something I missed here?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Also, I don't see that (at least on current Jenkins master branch) that Run.getLogFile is deprecated, which makes it feel very odd to me that we're deprecated its override in `WorkflowJob`

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: That would be the `FileLogStorage` branch. I.e., what is run if you do _not_ have the CloudWatch Logs plugin installed.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: > Isn't this potentially going to generate a log of logspam though?

Potentially. Note that the log call code was already there—not touched by this PR.

> I don't see that (at least on current Jenkins master branch) that Run.getLogFile is deprecated

Yup, on my to-do list.

svanoort: I think you might get away with marking it FINE or lower, but... as a user isn't it pretty obnoxious to see a bunch of log warnings due to a plugin using an API that isn't generally deprecated yet?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1080,16 +1081,21 @@ private void writeLogTo(WriteMethod method, OutputStream os) throws IOException
 
     @Override public File getLogFile() {
         LOGGER.log(Level.WARNING, "Avoid calling getLogFile on " + this, new UnsupportedOperationException());
-        try {
-            File f = File.createTempFile("deprecated", ".log", getRootDir());
-            f.deleteOnExit();
-            try (OutputStream os = new FileOutputStream(f)) {
-                getLogText().writeRawLogTo(0, os);
+        File f = super.getLogFile();
+        if (LogStorage.of(asFlowExecutionOwner()) instanceof FileLogStorage) {
```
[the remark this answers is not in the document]
svanoort: Note that this might be an escape hatch for us.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Yeah, but in that case... isn't that the *actual* build log?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Well, all warnings are obnoxious if (a) you look at them, (b) you are not in a position to fix them.

Anyway, again, that is not part of this PR. Can have a separate discussion about systems for limiting repeated log messages etc.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Gotta say it: we really, really, really should have a testcase covering these scenarios.  Relying on code review alone to catch issues here is not a longterm path for success.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
jglick: > returning a FileLogStorage even if there's no index file associated

Yup, and that is right. If there is no index file, `log` can still be read for the whole-log build. You will not get annotations about which lines came from particular steps, that is all.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: In the case of `FileLogStorage`, yes that is the actual build log, as it would be for freestyle builds.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
vivek: >>I don't see that (at least on current Jenkins master branch) that Run.getLogFile is deprecated

>Yup, on my to-do list.

Deprecation should happen first in upstream and then followed in downstream. In this case, we are doing opposite? Whats the rationale for that? If at all, this warning should be coming from upstream, no?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: 🐛 So we're marking for Jenkins to... delete our build's actual logfile on exit. 🐛

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: > first in upstream and then followed in downstream

In general, for JEP-210 the intention was to produce something that can run on existing LTS lines. There are various follow-up tasks. JEP-207 [discusses this particular topic](https://github.com/jenkinsci/jep/blob/master/jep/207/README.adoc#run-api-compatibility).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
svanoort: I'll put it this way: that feels like an *extremely surprising* outcome for someone interacting with that API.

jglick:
> concurrent calls to `getLogFile`

Yes, but the alternative would be the absence of this PR and a separate file per call, which is much worse than any such issue.

If you are a nondefault log storage, you really want to treat all calls to `getLogFile` as bugs and have them fixed, so any implementation is a best effort.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Ah, except... due to magic the legacy logs magically have become a FileLogStorage due to the other weirdness of the `LogStorage.of` call.  Which raises the question: what is this code addressing?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1080,16 +1081,21 @@ private void writeLogTo(WriteMethod method, OutputStream os) throws IOException
...
+            // FileLogStorage does write a file with the same name and the same format as before JEP-210, so accept it if it is there.
+        } else {
+            // Some other storage. Cache under the traditional name, but load from the defined source on each call.
```
svanoort: Question for @jglick - can you please enumerate in code comments which of these cases are meant to handle:

1. Traditional log storage on-master
2. The new log-rewrite log storage on-master
3. Cloud storage on-master

And I feel like we may actually be making a mistake by handing responsibility for those to the `WorkflowJob` or indeed the `Job` - it feels like the LogStorage itself ought to take ownership for these (including tests for each case).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Unless some plugin adds an alternative, `FileLogStorage` is in use. It does not matter if this build predates JEP-210.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
jglick: You should not be interacting with this API at all—it is deprecated—but if you do, it behaves exactly as it did before, and continues to do for freestyle builds: gives you a `File` containing the complete raw text (i.e., including encoded `ConsoleNote`s) of the build log.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: > can you please enumerate in code comments which of these cases are meant to handle

I already did, but I can try to reword the comments.

> * Traditional log storage on-master
> * The new log-rewrite log storage on-master

There is no difference between these two so far as the `log` file is concerned.

> 3\. Cloud storage on-master

There is no such thing.

> the LogStorage itself ought to take ownership for these

No. `LogStorage` handles writes to, and read from, logs wherever they may be stored, and defines the API for making that pluggable. Legacy code which presumes that the log is a file on the master has to be accommodated somehow, for compatibility.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1020,7 +1020,7 @@ public int hashCode() {
     @Override public InputStream getLogInputStream() throws IOException {
         // Inefficient but probably rarely used anyway.
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        writeLogTo(getLogText()::writeLogTo, baos);
+        writeLogTo(getLogText()::writeRawLogTo, baos);
```
dwnusbaum: Yikes, caused by my changes in #114. Good catch!

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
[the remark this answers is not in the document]
svanoort: I mean specifically the `LogStorage.of` API - agree that in an ideal world nobody would be trying to work with the file directly (although I bet it's still happening anyway)

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: What we should do is deprecate in Jenkins core (with an explanation in comments about why), log at a FINE level, and then when that releases in an LTS, we upgrade the log message to WARNING.  This also adds a window of time for any plugins that might be (ab)using getLogFile to change over before we starting making lots of logspam.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: > There is no such thing.

Heh, typing fast, but you know what I meant. :)

> No. LogStorage handles writes to, and read from, logs wherever they may be stored, and defines the API for making that pluggable

Per this: https://github.com/jenkinsci/jep/blob/master/jep/207/README.adoc#run-api-compatibility isn't the `LogStorage` method `File toLogFile()` supposed to be able to handle all of this? 

What, concretely, is the `else` case addressing here?  IIUC unless a key part was left un-implemented we might be able to remove most of this logic.

svanoort: And if there *is* a concrete case that we know the 'else' branch is covering, where is its testcase?

jglick: JEP-207 is only implemented in an experimental branch and I did not propose that API.

The `else` case will be active when you run the `pipeline-log-fluentd-cloudwatch` plugin.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
svanoort: Why?  While it may be satisfying to simple signal "stop using this API" by making it log a warning every time, for a *user* this looks like the Workflow Job plugin was broken by the update that added the logging - we've seen this before with "cosmetic" log warnings.  Just not great optics or user experience. 

Being able to point to deprecation of the extension point in core and an appropriate waiting period gives us more of a leg to stand on here.

jglick:
> the deprecation exceptions

I am certainly open to changes in logging, but this is really a separate topic. Ideally I would like to have something in Jenkins consolidate similar warnings—like a custom log handler in Winstone?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
timja: so what do I call instead?
I'm making changes to the build-failure-analyzer plugin, and saw my logs being spammed here.
but there's no javadoc explaining what the replacement / another solution is?

jglick: @timja Any log-related methods other than `getLogFile`. `getLogText` is the basic one. In your case I would suggest simply

```diff
-String currentFile = build.getLogFile().getName();
+String currentFile = "log";
```

timja: Will give it a go later on thanks 👍

jglick: > I don't see that (at least on current Jenkins master branch) that Run.getLogFile is deprecated

https://github.com/jenkinsci/jenkins/pull/3963

PR #122, opened 2019-01-22
Upgrade to workflow-support 3.1

bitwiseman:
Two tests began failing with this change.  Disabled for now.
Will review with team to see if they should be deleted or fixed.

Workflow-support 3.1 required updating to newer parent pom to satisfy enforcer.

Upgrade to JTH 2.46 not required, but previous version was over a year old.  Upgrade caused no breaks. 

@abayer @dwnusbaum  
I'm new enough in this area that I'm not sure if the failed tests are important and the linked bugs didn't cast much light on the failures.  Started to debug but chose to submit PR first for discussion.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.28</version>
+        <version>3.33</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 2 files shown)


bitwiseman:
Discussed with @abayer.  The test failures are correct.  This is a breaking changes that requires the local data to be upgraded.

abayer:
`loadMigratedBuildRecord` has a `program.dat`, so it's probably that. =)

dwnusbaum:
> loadMigratedBuildRecord has a program.dat, so it's probably that.

Yeah it's definitely related to the workflow-support 3.x upgrade, the question is just whether it is a general test intended to prevent accidental regressions in binary compatibility, in which case it could be worth recreating the test to help us going forward, or if it was just testing some one time migration in the distant past in which case the utility of recreating the test is probably not worth the effort.

dwnusbaum:
It would also be good to go ahead and update the rest of the workflow/pipeline/scm related plugins in the POM at the same time (unless any of them has a super high core baseline or something) to catch any other potential PCT issues, but also fine to leave them for a separate PR.

dwnusbaum:
Whoops, we should've merged this a long time ago, but it ended up getting subsumed by #126 to get the build passing on Java 11, so I'm closing it now.

PR #125, opened 2019-03-07
Java 11 readiness

batmat:
This fixes an issue surfaced through PCT runs.

@jenkinsci/java11-support FYI

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/MemoryCleanupTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/MemoryCleanupTest.java
@@ -12,6 +12,8 @@ import org.jvnet.hudson.test.MemoryAssert;
 import java.lang.ref.WeakReference;
 import java.util.logging.Level;
 
+import static org.junit.Assume.assumeTrue;
+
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/job/MemoryCleanupTest.java
```diff
@@ -24,6 +26,8 @@
 
     @Test
     public void cleanup() throws Exception {
+        // Ignore on Java 9+, because `MemoryAssert.assertGC` below tries to call setAccessible on some internal structure
+        assumeTrue(System.getProperty("java.specification.version").startsWith("1."));
```
MRamonLeon: WDYT about `assumeTrue(JavaSpecificationVersion.forCurrentJVM().isOlderThanOrEqualTo(JavaSpecificationVersion.JAVA_8));`
?

batmat: This line shouldn't be there actually. I was supposed to remove it in favor of the change directly in JTH.

PR #127, opened 2019-03-13
Fix prefix ending up outside of console

faircreek:
Pipeline prefixes how it looked in 2.25
pipeline_prefix_how_it_looks_in_2 25

How it looks in 2.32
pipeline_prefix_bug_2

This might be somewhat ok with short prefixes but with long ones it becomes hard to read.
pipeline_prefix_bug

It's a very small change to make it look like in 2.25.


dwnusbaum:
Thanks for the PR! See #111 for additional context around the change that added the current visualization after the log-handling rewrite removed the branches from the log itself. If we make this change I feel like it might make more sense to just move the branches back into the log itself so that they are also searchable via `ctrl+f` in most browsers. CC @jglick

faircreek:
Yep that is probably the easiest. In re to #111 that I didn't read before, I did actually figuring out what you tried to achieve.  Better is probably two "columns" one right adjusted for the prefixes and one left adjusted for the log. The whole thing does need to be in the console area though.  This is doable by calculating string lengths and the adjust the padding dynamically. The question is how it will look with very long prefixes, line wrapping and such.

jglick:
Rather than shifting the placement it may make more sense to trim long branch names and make the full version available via tooltip or similar.

Searchability is an independent question.

PR #130, opened 2019-05-16
Makes WorkflowRun.checkouts() public

mikeyeung123:
Allows plugins to access the SCMs in the current pipeline run. Currently the next best option is `WorkflowJob.getSCMs()`, which gets the SCMs of the _previous_ run.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -758,7 +758,8 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         return isBuilding(); // there is no equivalent to a post-production state for flows
     }
 
-    synchronized @Nonnull List<SCMCheckout> checkouts(@CheckForNull TaskListener listener) {
+    @Exported
+    public synchronized @Nonnull List<SCMCheckout> checkouts(@CheckForNull TaskListener listener) {
         if (checkouts == null) {
             LOGGER.log(Level.WARNING, "JENKINS-26761: no checkouts in {0}", this);
             if (listener != null) {


mikeyeung123:
I don't think the build error is because of the PR...

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -758,7 +758,8 @@ public boolean hasntStartedYet() {
         return isBuilding(); // there is no equivalent to a post-production state for flows
     }
 
-    synchronized @Nonnull List<SCMCheckout> checkouts(@CheckForNull TaskListener listener) {
+    @Exported
```
dwnusbaum: Does `@Exported` even work for a method that takes parameters?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -758,7 +758,8 @@ public boolean hasntStartedYet() {
...
+    public synchronized @Nonnull List<SCMCheckout> checkouts(@CheckForNull TaskListener listener) {
```
dwnusbaum: Since `SCMCheckout` is a package-private class, other plugins wouldn't be able to use this method without making that type public as well (You'd also need to add getters for its fields, etc.). I think it would be better to have a new method that returns `List<SCM>` (or maybe `List<? extends SCM>`, not sure what is normally used for this kind of thing) instead and populates that list using `WorkflowRun.checkouts` internally.

dwnusbaum:
> I don't think the build error is because of the PR...

You're right, it's not related, and it was fixed by updating to version 3.43 of the parent pom: https://github.com/jenkinsci/workflow-job-plugin/commit/76d44d5ea73ff763f86ad7b2c5f7fcac103eef07.

dwnusbaum:
The approach described in https://github.com/jenkinsci/workflow-job-plugin/pull/130#discussion_r295013836 was implemented by https://github.com/jenkinsci/workflow-job-plugin/pull/141, so I am closing this. If you need the information in the `SCMCheckout`s, not just the `SCM`, we'd need to either make that class public or think of another way to expose the desired information, but it would help to understand your use case before we go down that route.

PR #131, opened 2019-06-07
[JENKINS-46076] Reordering events so WorkflowRun.completed not set until finish cause printed & log closed

jglick:
[JENKINS-46076](https://issues.jenkins-ci.org/browse/JENKINS-46076)

After incremental deployment, downstream bumps in:
- [X] https://github.com/jenkinsci/workflow-cps-plugin/pull/294
- [X] https://github.com/jenkinsci/kubernetes-plugin/pull/496
- [X] https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/336

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -195,7 +195,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     /** True when first started, false when running after a restart. */
     private transient boolean firstTime;
 
-    /** Obtain our guard object for log copying, lazily initializing if needed.
+    /** Obtain our guard object for metadata, lazily initializing if needed.
      *  Note: to avoid deadlocks, when nesting locks we ALWAYS need to lock on the guard first, THEN the WorkflowRun.
      *  Synchronizing this helps ensure that fields are not mutated during a {@link #save()} operation, since that locks on the Run.
      */
@@ -211,8 +211,8 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
 
     /** Used internally to ensure listener has been initialized correctly. */
     BuildListener getListener() {
-        // Un-synchronized to prevent deadlocks (combination of run and logCopyGuard)
-        // Note that in portions where multithreaded access is possible we are already synchronizing on logCopyGuard
+        // Un-synchronized to prevent deadlocks (combination of run and metadataGuard)
+        // Note that in portions where multithreaded access is possible we are already synchronizing on metadataGuard
         if (listener == null) {
             try {
                 // TODO to better handle in-VM restart (e.g. in JenkinsRule), move CpsFlowExecution.suspendAll logic into a FlowExecution.notifyShutdown override, then make FlowExecutionOwner.notifyShutdown also overridable, which for WorkflowRun.Owner should listener.close() as needed
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -156,7 +156,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
             return WorkflowRun.this;
         }
     };
-    private transient BuildListener listener;
+    private transient volatile BuildListener listener;
 
     private transient boolean allowTerm;
 
@@ -210,7 +210,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     static final StreamBuildListener NULL_LISTENER = new StreamBuildListener(new NullStream());
 
     /** Used internally to ensure listener has been initialized correctly. */
-    BuildListener getListener() {
+    private BuildListener getListener() {
         // Un-synchronized to prevent deadlocks (combination of run and metadataGuard)
         // Note that in portions where multithreaded access is possible we are already synchronizing on metadataGuard
         if (listener == null) {
@@ -572,6 +572,9 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
     private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
         try {
             setResult(r);
+            synchronized (getMetadataGuard()) {
+                completed = true;
+            }
             duration = Math.max(0, System.currentTimeMillis() - getStartTimeInMillis());
             LOGGER.log(Level.INFO, "{0} completed: {1}", new Object[]{toString(), getResult()});
             if (listener == null) {
@@ -595,9 +598,6 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                     }
                 }
                 listener = null;
-            }
-            synchronized (getMetadataGuard()) {
-                completed = true;
             }
             saveWithoutFailing(); // TODO useless if we are inside a BulkChange
             Timer.get().submit(() -> {
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -570,16 +570,10 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
 
     /** Handles normal build completion (including errors) but also handles the case that the flow did not even start correctly, for example due to an error in {@link FlowExecution#start}. */
     private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
-        boolean nullListener = false;
-        synchronized (getLogCopyGuard()) {
-            nullListener = listener == null;
-            setResult(r);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -199,11 +199,11 @@ public String url() {
      *  Note: to avoid deadlocks, when nesting locks we ALWAYS need to lock on the logCopyGuard first, THEN the WorkflowRun.
```
bitwiseman:
```suggestion
     *  Note: to avoid deadlocks, when nesting locks we ALWAYS need to lock on this guard first, THEN the WorkflowRun.
```

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -570,16 +570,11 @@ private String key() {
 
     /** Handles normal build completion (including errors) but also handles the case that the flow did not even start correctly, for example due to an error in {@link FlowExecution#start}. */
     private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
-        boolean nullListener = false;
-        synchronized (getLogCopyGuard()) {
-            nullListener = listener == null;
+        try {
             setResult(r);
```
bitwiseman: Doesn't `setResult(r);` still need to be guarded?

jglick: I see no reason why it would.

bitwiseman:
https://github.com/jenkinsci/workflow-job-plugin/blob/fb2c9f5778e044e309ee887375fbd70b5f4cb872/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L214 - one last reference to "logCopyGuard"

bitwiseman:
https://github.com/jenkinsci/workflow-job-plugin/blob/fb2c9f5778e044e309ee887375fbd70b5f4cb872/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L214 - last reference to copyGuard  (I can't edit it or I'd just do it).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
bitwiseman: Other parts of this file have result setting inside the the synch block.

jglick: Only incidentally I think.

dwnusbaum: Once `result` is set to a non-null value, `isBuilding` will call [`isInProgress`](https://github.com/jenkinsci/workflow-job-plugin/blob/db692f18de005fe84d5a95b0bf74597fc8a201a8/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L746) instead of short-circuiting. Doesn't this mean that the behavior will be the same since `FlowExecution.isComplete() == true` (or is it false here?), so `isInProgress` just starts returning false out of the second branch rather than the first due to moving `completed` around?

A good downstream check might be to remove the workaround [here](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/31569c0385f555f0b3146c68eb748a5dcf16efd6/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java#L511-L539) and run pipeline-model-definition's tests a few times to see if we still run into related failures.

jglick: > Doesn't this mean

Unfortunately it does. Actually this could happen regardless of when we call `setResult` in `finish`, since this _may_ be called earlier on a truly running build (e.g., by `junit`). So I will need to do deeper fixes.

> remove the workaround [here](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/31569c0385f555f0b3146c68eb748a5dcf16efd6/pipeline-model-definition/src/test/java/org/jenkinsci/plugins/pipeline/modeldefinition/AbstractModelDefTest.java#L511-L539)

Thanks for tip, I did not know about this one.

jglick: I checked the comment in `pipeline-model-definition`. I believe that situation is unaffected by this patch:

* `waitForCompletion` starts waiting
* `result` is set to something in the middle of the build, say `UNSTABLE`
* `FlowEndNode` is added to the graph
* (asynch) `GraphListener`s start to be notified
* `isBuilding` is checked
  * `result == null` is false, so `isInProgress()` is called
  * `completed` is still false
  * `exec.isComplete()` is true
* so `assertBuildStatus` fails
* now `GraphL` is notified and `finish` is called
  * in some order, `result` is updated, `completed` is set to `true`, etc.

Fixing that kind of failure would require a more intrusive patch to `isBuilding` which is not required for the originally reported class of flakes and which I am not attempting here.

jglick:
Now waiting for [INFRA-2140](https://issues.jenkins-ci.org/browse/INFRA-2140).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: …though you could probably do something like

```java
j.waitForCompletion(run);
j.waitForMessage("Finished: FAILURE", run);
```

which would be reliable after this patch.

Not going to try it so long as https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/325 remains unmerged.

dwnusbaum: > Fixing that kind of failure would require a more intrusive patch to isBuilding which is not required for the originally reported class of flakes and which I am not attempting here

Fair enough. My first thought is to add a `transient volatile boolean currentlyCompleting` field, set it to true before calling `setResult()` or setting `completed = true` in `finish` (set it back to false in a finally block before notifying listeners in `finish`), and then in `isInProgress`, if that field is true, return true (if `finish` can be called multiple times, you'd need something more complicated so a finished build couldn't go back to being in-progress, although it seems unlikely that anyone would be polling `isBuilding` in that case).

jglick: Would not work unless you registered a `GraphListener.Synchronous` which set this field.

Simpler would be to make `isBuilding` check, say, `listener == null` like `isLogUpdated` now does, but I am not confident in such a change. Of course all these methods are only scantily documented in `Run`, since their original meaning was heavily tied to `AbstractBuild` semantics.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -601,6 +596,9 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
                 }
                 listener = null;
             }
+            synchronized (getMetadataGuard()) {
+                completed = true;
```
dwnusbaum: Moving this around doesn't actually matter as discussed in https://github.com/jenkinsci/workflow-job-plugin/pull/131/files#r291657569, right? If not, I'd just move it back to where it was originally.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -755,7 +753,7 @@ public boolean hasntStartedYet() {
     }
 
     @Override public boolean isLogUpdated() {
-        return isBuilding(); // there is no equivalent to a post-production state for flows
+        return listener != null || isBuilding(); // there is no equivalent to a post-production state for flows
```
dwnusbaum: IIUC, this is the main fix. Does `listener` need to be made `volatile` now to make sure that tests calling `assertLogContains` actually see the latest value?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Mm, can try it. Means going back through a bunch of rounds of integration testing, unfortunately.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Maybe. I doubt any of our tests could observe anything close to the time resolution required for `volatile` to matter. Usually it is added to fields to ensure that observers do not see half-constructed objects, which is not relevant here since we are merely checking for nullness.

jglick:
Seems to have worked—all downstream builds pass.

Vlatombe:
Hey @dwnusbaum, would you mind running a release of the plugin? I'd like to integrate the changes in this PR in downstream plugins

dwnusbaum:
@Vlatombe Ok, I will try to release the plugin today.

Vlatombe:
cool, thanks!

jglick:
I suppose you can just suggest changes in the downstream POMs after it has been released.

dwnusbaum:
2.33 was released, I added suggestion to the downstream PRs linked here.

PR #132, opened 2019-07-20
Use Functions.printStackTrace

basil:
In #38, a call to `Functions.printThrowable` was added with a TODO comment to switch to `Functions.printStackTrace` when the baseline was bumped to 2.43+. The baseline is now 2.121.1, so this TODO can be completed.

This code doesn't appear to have any test coverage, but I feel confident about the change based on the similar logic that is present in `WorkflowRun` and the testing I did for a similar change in jenkinsci/workflow-basic-steps-plugin#91.


basil:
Re-triggering build.

PR #133, opened 2019-07-20
Consistently use sandbox for CpsFlowDefinitions

basil:
While looking at the test suite for this plugin, I noticed the CpsFlowDefinitions in the tests don't consistently use the script security sandbox. Always using the script security sandbox makes the tests consistent with each other, and it's also a more realistic environment given that the script security sandbox should always be enabled in production.

I did the following in this change:

- In any case where the deprecated single-argument constructor for `CpsFlowDefinition` was used, I switched to the non-deprecated two-argument constructor (which takes a boolean as the second argument to determine whether to enable the script security sandbox).
- Wherever the script security sandbox could be enabled, I enabled it.
- Wherever the script security sandbox couldn't be enabled, I added a comment explaining why. This seemed preferable to approving all the relevant methods the test used, which could be fragile since many of those methods are implementation details of classes outside of this project.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
@@ -92,7 +92,7 @@ public class WorkflowRunTest {
 
     @Test public void basics() throws Exception {
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
-        p.setDefinition(new CpsFlowDefinition("println('hello')"));
+        p.setDefinition(new CpsFlowDefinition("println('hello')", true));
         WorkflowRun b1 = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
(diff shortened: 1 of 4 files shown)


src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -227,7 +227,7 @@ private void assertColor(WorkflowRun b, BallColor color) throws IOException {
         final WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
         final String groovy = "println 'hello'";
         try (ACLContext context = ACL.as(User.getById("dev", true))) {
-            p.setDefinition(new CpsFlowDefinition(groovy));
+            p.setDefinition(new CpsFlowDefinition(groovy, false /* for mock authorization strategy */));
```
dwnusbaum: Maybe clearer to use an explanation like "for full script approval with non-privileged user"

PR #136, opened 2019-08-19
Trivial code cleanup

basil:
A series of trivial code cleanup changes for this repository. Most of these changes remove usages of deprecated functionality or otherwise fix IDE warnings. I kept each commit separate. Each commit message should be self-explanatory, but if anything is unclear I am happy to provide further explanation. Also, feel free to revert any specific cleanup commits that are unwanted.


src/main/java/org/jenkinsci/plugins/workflow/job/views/GraphVizAction.java
```diff
@@ -74,7 +74,7 @@ private GraphVizAction(WorkflowRun run) {
     public HttpResponse doDot() throws IOException {
         StringWriter sw = new StringWriter();
         writeDot(new PrintWriter(sw));
-        return HttpResponses.plainText(sw.toString());
+        return HttpResponses.text(sw.toString());
```
dwnusbaum: So IIUC, the difference is that the new method flushes instead of appending a newline? Did you check via the UI to make sure the output still looks ok? It's probably fine, but good to be careful.

src/test/java/org/jenkinsci/plugins/workflow/job/CpsPersistenceTest.java
```diff
@@ -197,10 +198,10 @@ public void completedFinalFlowNodeNotPersisted() throws Exception {
             String finalId = run.getExecution().getCurrentHeads().get(0).getId();
 
             // Hack but deletes the file from disk
-            CpsFlowExecution cpsExec = (CpsFlowExecution)(run.getExecution());
+            CpsFlowExecution cpsExec = (CpsFlowExecution) run.getExecution();
             File f = new File(cpsExec.getStorageDir(), finalId+".xml");
-            f.delete();
-            executionAndBuildResult[0] = ((CpsFlowExecution)(run.getExecution())).getResult();
+            Files.delete(f.toPath());
```
dwnusbaum: This will throw an exception if the file doesn't exist where the old code wouldn't have, but I think that is a good thing, since it keeps us from unintentionally testing a different scenario or potentially having a test pass for the wrong reason.

src/test/java/org/jenkinsci/plugins/workflow/job/CpsPersistenceTest.java
```diff
@@ -336,17 +337,17 @@ public void inProgressButFlowNodesLost() throws Exception {
...
         story.thenWithHardShutdown( j -> {
             WorkflowRun run = runBasicPauseOnInput(j, DEFAULT_JOBNAME, build);
-            CpsFlowExecution cpsExec = (CpsFlowExecution)(run.getExecution());
+            CpsFlowExecution cpsExec = (CpsFlowExecution) run.getExecution();
             Method m = cpsExec.getClass().getDeclaredMethod("getProgramDataFile", null);
             m.setAccessible(true);
-            File f = (File)(m.invoke(cpsExec, null));
-            f.delete();
+            File f = (File) m.invoke(cpsExec, null);
+            Files.delete(f.toPath());
```
dwnusbaum: ditto

src/main/java/org/jenkinsci/plugins/workflow/job/views/GraphVizAction.java
basil: > So IIUC, the difference is that the new method flushes instead of appending a newline?

Yeah, that is my understanding as well.

> Did you check via the UI to make sure the output still looks ok? It's probably fine, but good to be careful.

I added `r.interactiveBreak();` to the end of `WorkflowRunTest#culprits`, went to `http://${JENKINS_URL}/jenkins/job/p/2/graphViz/`, and got [this image](https://i.ibb.co/bBwkg6d/graphViz.png).

dwnusbaum: Great, thanks for checking!

PR #138, opened 2019-08-27
Adding LICENSE file to project

kshithijiyer:


kshithijiyer:
It is good to have a dedicated License file. Not every project will have a pom.xml and when someone is looking for a place to find what license a project is open sourced under than he'll generally look for a license file.

kshithijiyer:
The intent here is to have a general license file common to all the code and each author can have their own license added below that one in the same file. So if someone what's the know the licenses of the code he/she can see the license file which will state Licenses for all the code in a repo.

LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017 CloudBees, Inc.
```
jglick, line 3:
```suggestion
Copyright (c) 2022 CloudBees, Inc. and other contributors.
```

jglick (changes requested):

PR #141, opened 2019-08-28
Add public method to retrieve list of SCMs checked out

car-roll:
Making available all the `SCM` Objects used in the `checkouts` field.

CC  @dwnusbaum

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -774,7 +774,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         return checkouts;
     }
 
-    public @Nonnull List<SCM> getCheckouts() {
+    public @Nonnull List<SCM> getSCMs() {
         List<SCMCheckout> scmCheckouts = checkouts(TaskListener.NULL);
         List<SCM> scmList = new ArrayList<>();
         for (SCMCheckout checkout : scmCheckouts) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -775,15 +775,13 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         return checkouts;
     }
 
-    public synchronized @Nullable List<SCM> getCheckouts() {
-       if (checkouts == null) {
-           return null;
-       }
-       List<SCM> scmList = new ArrayList<>();
-       for (SCMCheckout aCheckout : checkouts) {
-           scmList.add(aCheckout.getScm());
-       }
-       return scmList;
+    public @Nonnull List<SCM> getCheckouts() {
+        List<SCMCheckout> scmCheckouts = checkouts(TaskListener.NULL);
+        List<SCM> scmList = new ArrayList<>();
+        for (SCMCheckout checkout : scmCheckouts) {
+            scmList.add(checkout.getScm());
+        }
+        return scmList;
     }
 
     @Override
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -85,6 +85,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import jenkins.model.CauseOfInterruption;
 import jenkins.model.Jenkins;
 import jenkins.model.lazy.BuildReference;
@@ -774,6 +775,17 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
         return checkouts;
     }
 
+    public synchronized @Nullable List<SCM> getCheckouts() {
+       if (checkouts == null) {
+           return null;
+       }
+       List<SCM> scmList = new ArrayList<>();
+       for (SCMCheckout aCheckout : checkouts) {
+           scmList.add(aCheckout.getScm());
+       }
+       return scmList;
+    }
+
     @Override
     @Exported
     public synchronized List<ChangeLogSet<? extends ChangeLogSet.Entry>> getChangeSets() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -771,6 +772,17 @@ public boolean hasntStartedYet() {
         return checkouts;
     }
 
+    public synchronized @Nullable List<SCM> getCheckouts() {
+       if (checkouts == null) {
```
dwnusbaum: Instead of having this method be `synchronized` and using the `checkouts` field directly, I think you should use `checkouts(TaskListener.null)` instead, then store the result in a local variable and use that in this method so that the Javadoc comment on `checkouts` that says `Access only through checkouts(TaskListener)` stays true and we don't have to worry about synchronization on the same object from multiple methods (even though in this case it is easy to reason about for now).

If you do that, the return value here would be `@Nonnull` since `checkouts(TaskListener)` never returns null, and that seems easier for consumers to handle.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -85,6 +85,7 @@
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
```
dwnusbaum: nit: you could remove this now:

```suggestion
```

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -774,6 +774,15 @@ public boolean hasntStartedYet() {
         return checkouts;
     }
 
+    public @Nonnull List<SCM> getCheckouts() {
```
dwnusbaum:
```suggestion
    public @Nonnull List<SCM> getSCMs() {
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -473,4 +479,17 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
         assertFalse(b.isLogUpdated());
     }
 
+    @Test public void getScms() throws Exception {
+        sampleRepo.init();
+        sampleRepo.write("Jenkinsfile", "echo 'hello'");
+        sampleRepo.git("add", "Jenkinsfile");
+        sampleRepo.git("commit", "--message=files");
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        CpsScmFlowDefinition def = new CpsScmFlowDefinition(new GitStep(sampleRepo.toString()).createSCM(), "Jenkinsfile");
```
dwnusbaum: nit: No need to go through `GitStep` like this, I think you can just create a `GitSCM` directly:

```suggestion
        CpsScmFlowDefinition def = new CpsScmFlowDefinition(new GitSCM(sampleRepo.toString()), "Jenkinsfile");
```

(and then remove the GitStep import)

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -473,4 +479,17 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+        p.setDefinition(def);
+        WorkflowRun b = r.buildAndAssertSuccess(p);
+        List<SCM> checkouts = b.getCheckouts();
```
dwnusbaum:
```suggestion
        List<SCM> checkouts = b.getSCMs();
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -52,6 +54,8 @@
 import jenkins.model.CauseOfInterruption;
 import jenkins.model.InterruptedBuildAction;
 import jenkins.model.Jenkins;
+import jenkins.plugins.git.GitSampleRepoRule;
+import jenkins.plugins.git.GitStep;
```
dwnusbaum:
```suggestion
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -473,4 +479,17 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+        sampleRepo.write("Jenkinsfile", "echo 'hello'");
+        sampleRepo.git("add", "Jenkinsfile");
+        sampleRepo.git("commit", "--message=files");
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        CpsScmFlowDefinition def = new CpsScmFlowDefinition(new GitStep(sampleRepo.toString()).createSCM(), "Jenkinsfile");
+        p.setDefinition(def);
+        WorkflowRun b = r.buildAndAssertSuccess(p);
+        List<SCM> checkouts = b.getCheckouts();
+        assertTrue(checkouts.size() == 1);
+        assertTrue(checkouts.get(0).getClass().getName().equals(GitSCM.class.getName()));
```
dwnusbaum: nit: Always better to use `assertEquals` or `assertThat` for better messages if the assertion fails.

```suggestion
        assertEquals(GitSCM.class, checkouts.get(0).getClass());
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -473,4 +479,17 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+        sampleRepo.init();
+        sampleRepo.write("Jenkinsfile", "echo 'hello'");
+        sampleRepo.git("add", "Jenkinsfile");
+        sampleRepo.git("commit", "--message=files");
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        CpsScmFlowDefinition def = new CpsScmFlowDefinition(new GitStep(sampleRepo.toString()).createSCM(), "Jenkinsfile");
+        p.setDefinition(def);
+        WorkflowRun b = r.buildAndAssertSuccess(p);
+        List<SCM> checkouts = b.getCheckouts();
+        assertTrue(checkouts.size() == 1);
```
dwnusbaum:
```suggestion
        assertEquals(1, checkouts.size());
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
car-roll: it had been a while that i forgot which one was the preferred route :p

dwnusbaum: No worries, with `assertTrue`/`assertFalse`, the failure just says "assertion failed", but with `assertEquals`, the message will show you the expected and actual values, which is much more useful. If all you have is a boolean value, `assertTrue`/`assertFalse` are fine, but if you are comparing values, better to use `assertEquals` or `assertThat` for complex comparisons.

PR #142, opened 2019-10-04
FlowExecutionListener fireCompleted before logs are closed

joseblas:
When writing to logs within *FlowExecutionListener.onCompleted* using this:

`owner.getListener().getLogger().println("blah blah blah");`

and exception is raised

java.lang.IllegalStateException: trying to open a build log on timeout-too-long-with-error #1 after it has completed
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.getListener(WorkflowRun.java:219)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.access$300(WorkflowRun.java:133)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun$Owner.getListener(WorkflowRun.java:955)
        at myListener(...)

In order to fix that, 

`FlowExecutionListener.fireCompleted(exec);`

should be executed before closing the stream and finishing the execution.

In this PR, the execution happens before closing the listener.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -593,6 +593,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                     Functions.printStackTrace(t, getListener().getLogger());
                 }
                 getListener().finished(getResult());
+                fireBeforeCompletedFlowExecutionListener();
                 if (listener instanceof AutoCloseable) {
                     try {
                         ((AutoCloseable) listener).close();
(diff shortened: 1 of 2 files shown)


abayer:
Doesn't seem unreasonable to me. Would like @dwnusbaum's review, of course. =)

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -593,6 +593,7 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
                     Functions.printStackTrace(t, getListener().getLogger());
                 }
                 getListener().finished(getResult());
+                fireBeforeCompletedFlowExecutionListener();
```
dwnusbaum: Looking at this some more, I think the key thing for `run.isCompleted()` to return true is the earlier line that sets `completed = true`, so I think `fireBeforeCompleted` and `fireCompleted` as currently written would have the same behavior other than the listener being closed in `fireCompleted`. In that case, given that we have `RunListener.fireCompleted` a few lines up, it seems like it would be reasonable to move the call `FlowExecutionListener.fireCompleted(exec)` next to it, in which case we wouldn't need a new method.

Did you do any testing to check what differences there would be between `fireBeforeCompleted` and `fireCompleted`? I am just guessing here from reading the code, but it would be good to add some tests if we do need a new method, pointing out the key differences between them. If just moving `fireCompleted` around works, then I would add a test using `fireCompleted` in a way that wouldn't have worked before it was moved.

dwnusbaum: Also FWIW, the call to `fireCompleted` appears to have been added in https://github.com/jenkinsci/workflow-job-plugin/pull/51, and just put directly at the end of the method, so it doesn't look like we would be undoing any previous fixes by moving it earlier in the method or anything. I also checked GitHub for any uses of `FlowExecutionListener.onCompleted` to see if they were doing anything that might break. I found one in [pipeline-config-history](https://github.com/jenkinsci/pipeline-config-history-plugin/blob/5f634be467af27b2eb5f778d8f95c76d21623066/src/main/java/org/jenkinsci/plugins/pipelineConfigHistory/model/PipelineConfigHistoryFlowExecutionListener.java#L59), and that looks like it would be fine even if the call was moved earlier.

PR #143, opened 2019-10-09
Changing WorkflowRun in order to run FlowExecutionListener.fireComple…

joseblas:
…ted just after unListener.fireCompleted and before closing the listener

Added some extra work in the test to make sure that some actions are allowed

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunWithFlowExecutionListenerTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunWithFlowExecutionListenerTest.java
@@ -1,0 +1,95 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2014, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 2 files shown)


joseblas:
related to https://github.com/jenkinsci/workflow-job-plugin/pull/142/files

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunWithFlowExecutionListenerTest.java
```diff
@@ -0,0 +1,114 @@
...
+        public void onCompleted(@Nonnull FlowExecution execution) {
+            super.onCompleted(execution);
+            Queue.Executable executable = null;
+            try {
+                PrintStream logger = execution.getOwner().getListener().getLogger();
+                logger.println("blah blah blah");
+                executable = execution.getOwner().getExecutable();
+                if( executable instanceof WorkflowRun){
+                    WorkflowRun run = (WorkflowRun) executable;
+                    Action action = new Action() {
```
dwnusbaum: nit: It doesn't look like the test uses this action anywhere, and it doesn't seem related to when the listener fires, so I'd just delete it to keep the test simple.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunWithFlowExecutionListenerTest.java
```diff
@@ -0,0 +1,114 @@
...
+                            return null;
+                        }
+                    };
+                    run.addAction(action);
+                    WorkflowJob workflowJob = run.getParent();
+                    int number = workflowJob.getLastBuild().number;
+                    logger.println("Build Number :" + number);
+                }
+            } catch (IOException e) {
+                e.printStackTrace();
```
dwnusbaum: I would rethrow this as an unchecked exception instead so that the build and test would fail if there was a problem:

```suggestion
                throw new RuntimeException(e);
```

PR #146, opened 2019-11-21
Do not print FlowDurabilityHint to build log

jglick:
I suspect few people ever override this setting, or even know what it means, so printing

```
Running in Durability level: MAX_SURVIVABILITY
```

at the start of every build log is just noise. (Also it is using a Java enum name, which is not very friendly.)


jglick:
> otherwise I don't think there is an easy way for them to see what setting is being used in a particular build

Maybe it should print a message only when the setting differs from the global default? Of course the global default may be changed over time etc.

I feel like this is something better left to system logs, not the build log.

dwnusbaum:
> Maybe it should print a message only when the setting differs from the global default? Of course the global default may be changed over time etc.

Yeah IDK. I was thinking maybe something like an `Action` on `WorkflowRun` with a `summary.jelly` to show the durability level used for the build on the main build page, but that's still a bit much.

car-roll (approved):

bitwiseman (approved):

PR #150, opened 2020-01-27
[JENKINS-60862] Add hidden spans with node data to truncated logs so branch names can be displayed

dwnusbaum:
See [JENKINS-60862](https://issues.jenkins-ci.org/browse/JENKINS-60862).

I am opening this PR as a draft to get feedback on the overall approach and to see if anyone has any ideas on a better way to fix the issue.

When viewing truncated logs, this PR directly injects `<span>` elements inside of a hidden `<div>`  for all flow nodes containing the same data that would be added for the `NewNodeConsoleNote` for that node if it was on screen, so that `NewNodeConsoleNote/script.js` always has all of the data it needs to be able to display branch names.

Here are screenshots showing the effect of the change:
<summary>Before (no branch names because the branches started before the truncation point):
<details>
Screen Shot 2020-01-27 at 14 45 37
</details>
</summary>
<summary>After (branch names shown):
<details>
Screen Shot 2020-01-27 at 14 44 47
</details>
</summary>

This approach seems pretty awkward, and requires us to clone `console.jelly` from core. One improvement we could make while keeping the same general approach would be to modify `FileLogStorage` (and any other implementations of `LogStorage`) to directly inject the same content into the log stream inside of `writeHtmlTo` in the `AnnotatedLargeText` override in `overallLog` whenever the offset is non-zero, so that we don't need to clone `console.jelly` or modify `WorkflowRun`. I just kept things simple for now and went with the one-plugin fix to get feedback on the overall idea.

I'm not sure about adding an automated test for this. I will check if there is a way to see what CSS rules are enabled on a page using HtmlUnit in a way that supports CSS rules injected via JavaScript.


kshultzCB:
Implementation aside, I like the idea of being able to see the branch names that way.

dwnusbaum:
@kshultzCB Just to clarify, branch names already show up like that, they just don't always show up correctly when you are viewing truncated logs instead of the full logs.

kshultzCB:
> @kshultzCB Just to clarify, branch names already show up like that, they just don't always show up correctly when you are viewing truncated logs instead of the full logs.

That's the thing I really like - that I don't have to load that entire mile-long full log to see the stage names. That would be a nice usability improvement.

bitwiseman:
This works fine, I'm happy with a single plugin change especially in the short to medium term. 

How hard would it be to switch to the other approach later?

dwnusbaum:
> How hard would it be to switch to the other approach later?

Taking this approach now shouldn't make it any harder to switch to any other approach later. Switching to injecting the spans into the log stream over in workflow-api would be relatively straightforward, we'd just move the logic over there, then remove it from this plugin and update the workflow-api dependency, so users wouldn't have to do anything special.

jessejoe:
FYI this change seems to have broken `/console` and `/consoleFull` for some users with large `parallel` pipeline logs (any version of this plugin with these changes is unusable for us). I've documented a lot of what I've found with a very simple repro case in https://issues.jenkins.io/browse/JENKINS-62241. We've also been discussing in the Jenkins Gitter channel.

jglick:
#164?

jessejoe:
@jglick thanks, I wish I'd found that a few days ago!

PR #152, opened 2020-02-28
[JENKINS-61415] Remove polling baselines from Pipelines when the SCM has no polling baseline

alexanderrtaylor:
If you check the changelog or polling  checkbox or make them true, then do a build it will make those checkout endpoints poll-able. But once you uncheck those or mark them false it never takes those endpoints off the possible endpoints to check.

This PR fixes that

It has tests included to prove when and where this works and doesnt work. Eventually I want a better fix as this will just mark it as "no polling baseline" which is not an accurate message

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -591,6 +591,7 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements L
                 FilePath workspace;
                 Launcher launcher;
                 WorkspaceList.Lease lease;
+
                 if (co.scm.requiresWorkspaceForPolling()) {
                     Jenkins j = Jenkins.getInstanceOrNull();
                     if (j == null) {
@@ -634,12 +635,22 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements L
     }
     @Extension public static final class SCMListenerImpl extends SCMListener {
         @Override public void onCheckout(Run<?,?> build, SCM scm, FilePath workspace, TaskListener listener, File changelogFile, SCMRevisionState pollingBaseline) throws Exception {
-            if (build instanceof WorkflowRun && pollingBaseline != null) {
+            if (build instanceof WorkflowRun) {
                 WorkflowJob job = ((WorkflowRun) build).getParent();
-                if (job.pollingBaselines == null) {
-                    job.pollingBaselines = new ConcurrentHashMap<>();
+                if (pollingBaseline != null) {
+                    if (job.pollingBaselines == null) {
+                        job.pollingBaselines = new ConcurrentHashMap<>();
+                    }
+                    job.pollingBaselines.put(scm.getKey(), pollingBaseline);
                 }
-                job.pollingBaselines.put(scm.getKey(), pollingBaseline);
+                else{
+                    //pollingBaseline is null here
+                    if (job.pollingBaselines != null) {
+                        // We need to remove the polling baseline if the baseline goes back to null
+                        // this is caused by the checkout/polling values becoming false
+                        job.pollingBaselines.remove(scm.getKey());
+                    }
+                }
             }
         }
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -642,8 +642,7 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements L
                         job.pollingBaselines = new ConcurrentHashMap<>();
                     }
                     job.pollingBaselines.put(scm.getKey(), pollingBaseline);
-                }
-                else{
+                } else {
                     //pollingBaseline is null here
                     if (job.pollingBaselines != null) {
                         // We need to remove the polling baseline if the baseline goes back to null
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -386,7 +386,7 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements L
      */
     @Deprecated
     public static final Permission ABORT = CANCEL;
-    
+
     @Override public Collection<? extends SubTask> getSubTasks() {
         // TODO mostly copied from AbstractProject, except SubTaskContributor is not available:
         List<SubTask> subTasks = new ArrayList<>();
@@ -591,7 +591,6 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements L
                 FilePath workspace;
                 Launcher launcher;
                 WorkspaceList.Lease lease;
-
                 if (co.scm.requiresWorkspaceForPolling()) {
                     Jenkins j = Jenkins.getInstanceOrNull();
                     if (j == null) {
@@ -637,19 +636,10 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements L
         @Override public void onCheckout(Run<?,?> build, SCM scm, FilePath workspace, TaskListener listener, File changelogFile, SCMRevisionState pollingBaseline) throws Exception {
             if (build instanceof WorkflowRun) {
                 WorkflowJob job = ((WorkflowRun) build).getParent();
-                if (pollingBaseline != null) {
-                    if (job.pollingBaselines == null) {
-                        job.pollingBaselines = new ConcurrentHashMap<>();
-                    }
-                    job.pollingBaselines.put(scm.getKey(), pollingBaseline);
-                } else {
-                    //pollingBaseline is null here
-                    if (job.pollingBaselines != null) {
-                        // We need to remove the polling baseline if the baseline goes back to null
-                        // this is caused by the checkout/polling values becoming false
-                        job.pollingBaselines.remove(scm.getKey());
-                    }
+                if (job.pollingBaselines == null) {
+                    job.pollingBaselines = new ConcurrentHashMap<>();
                 }
+                job.pollingBaselines.compute(scm.getKey(), (key, oldBaseline) -> pollingBaseline);
             }
         }
     }
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
@@ -72,6 +72,8 @@ import org.jenkinsci.plugins.workflow.graph.FlowGraphWalker;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.graph.StepNode;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
+
+import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.*;
 import org.junit.ClassRule;
 import org.junit.Rule;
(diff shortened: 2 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -491,4 +495,202 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+    // Basic test to prove that checkout gets turned off correctly
+    public void baselineResetSingleRepo() throws Exception {
+        sampleRepo.init();
+        TaskListener listener = StreamTaskListener.fromStdout();
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("semaphore 'waitFirst'\n" +
+                "node {\n" +
+                "    checkout(changelog:" + true +", poll:" + true + ", scm: [$class: 'GitSCM', branches: [[name: '*/master']], " +
+                "doGenerateSubmoduleConfigurations: false, extensions: [], " +
+                "gitTool: 'Default', submoduleCfg: [], userRemoteConfigs: [[url: '" + sampleRepo + "']]])" +
```
bitwiseman: `sampleRepo` ... the string passed to `url:` needs some modification to work on Windows.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -491,4 +495,202 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+        r.assertBuildStatusSuccess(r.waitForCompletion(b4));
+        for (WorkflowRun.SCMCheckout co : b4.checkouts(listener)){
+            assertTrue(co.pollingBaseline == null);
+        }
+    }
+
+    @Test
+    @Issue("NGPIPELINE-917")
+    // Basic test to prove that checkout gets individually set for each checkout
+    public void baselineResetMultipleRepos() throws Exception {
```
bitwiseman: What happens if you checkout the same repo twice in one pipeline with different settings?

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -491,4 +495,202 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+        p.setDefinition(new CpsFlowDefinition("semaphore 'waitFirst'\n" +
+                "node {\n" +
+                "    checkout(changelog:" + true +", poll:" + true + ", scm: [$class: 'GitSCM', branches: [[name: '*/master']], " +
+                "doGenerateSubmoduleConfigurations: false, extensions: [], " +
+                "gitTool: 'Default', submoduleCfg: [], userRemoteConfigs: [[url: '" + sampleRepo + "']]])" +
+                "}\n", false /* for org.jvnet.hudson.test.FakeChangeLogSCM */));
+
+        WorkflowRun b1 = p.scheduleBuild2(0).waitForStart();
+
+        SemaphoreStep.waitForStart("waitFirst/1", b1);
```
bitwiseman: Is this redundant given `.waitForStart()` above?

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -491,4 +495,202 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+            System.out.println("Checkout: " + co.toString());
+            assertTrue(co.pollingBaseline != null);
+        }
+
+        p.setDefinition(new CpsFlowDefinition("semaphore 'waitFirst'\n" +
+                "node {\n" +
+                "    checkout(changelog:" + false +", poll:" + false + ", scm: [$class: 'GitSCM', branches: [[name: '*/master']], " +
+                "doGenerateSubmoduleConfigurations: false, extensions: [], " +
+                "gitTool: 'Default', submoduleCfg: [], userRemoteConfigs: [[url: '" + sampleRepo + "']]])" +
+                "}\n", false /* for org.jvnet.hudson.test.FakeChangeLogSCM */));
```
bitwiseman: You do this same thing repeatedly. Wrap it in a method.

bitwiseman:
```suggestion
        setTestPipelineDefinition(sampleRepo, false, false);
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -491,4 +495,202 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+        for (WorkflowRun.SCMCheckout co : b1.checkouts(listener)){
+            System.out.println("Checkout: " + co.toString());
+            assertTrue(co.pollingBaseline != null);
+        }
+
+        p.setDefinition(new CpsFlowDefinition("semaphore 'waitFirst'\n" +
+                "node {\n" +
+                "    checkout(changelog:" + false +", poll:" + false + ", scm: [$class: 'GitSCM', branches: [[name: '*/master']], " +
+                "doGenerateSubmoduleConfigurations: false, extensions: [], " +
+                "gitTool: 'Default', submoduleCfg: [], userRemoteConfigs: [[url: '" + sampleRepo + "']]])" +
```
bitwiseman:
```suggestion
                "node {\n" +
                pollingTestCheckoutStep(sampleRepo, false, false) +
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
dwnusbaum: They have different meanings (the first one waits for the job to enter the queue and start running, the second waits for the corresponding `semaphore` step in the Pipeline to start executing), but in this case, since the test doesn't do or check anything between `SemaphoreStep.waitForStart` and `SemaphoreStep.success`, the semaphore step is redundant and it could just be removed, but so is the `.waitForStart()` above, since the next line just waits for the build to complete. This should be all that is needed:

```
WorkflowRun b1 = r.buildAndAssertSuccess(p);
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -491,4 +495,202 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+
+        WorkflowRun b2 = p.scheduleBuild2(0).waitForStart();
+
+        SemaphoreStep.waitForStart("waitFirst/2", b2);
+        SemaphoreStep.success("waitFirst/2", null);
+
+        r.assertBuildStatusSuccess(r.waitForCompletion(b2));
+        for (WorkflowRun.SCMCheckout co : b2.checkouts(listener)){
+            assertTrue(co.pollingBaseline != null);
+        }
```
bitwiseman:
```suggestion
        assertThat(b2.checkouts(listener).size(), equalTo(1);
        assertThat(b2.checkouts(listener).get(1).pollingBaseline, NonNullValue());
```

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -634,12 +635,22 @@ public void replaceAction(Action a) {
     }
     @Extension public static final class SCMListenerImpl extends SCMListener {
         @Override public void onCheckout(Run<?,?> build, SCM scm, FilePath workspace, TaskListener listener, File changelogFile, SCMRevisionState pollingBaseline) throws Exception {
-            if (build instanceof WorkflowRun && pollingBaseline != null) {
+            if (build instanceof WorkflowRun) {
                 WorkflowJob job = ((WorkflowRun) build).getParent();
```
bitwiseman: Can `getParent()` ever return null?

bitwiseman: Duh.  No.  Silly me.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -634,12 +635,22 @@ public void replaceAction(Action a) {
...
-                if (job.pollingBaselines == null) {
-                    job.pollingBaselines = new ConcurrentHashMap<>();
+                if (pollingBaseline != null) {
+                    if (job.pollingBaselines == null) {
+                        job.pollingBaselines = new ConcurrentHashMap<>();
+                    }
+                    job.pollingBaselines.put(scm.getKey(), pollingBaseline);
+                }
+                else{
```
bitwiseman:
```suggestion
                } else {
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -491,4 +498,208 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+        r.assertBuildStatusSuccess(r.waitForCompletion(b2));
+        assertThat(b2.checkouts(listener).size(), equalTo(1));
+        //proving the baseline is removed
+        assertThat(b2.checkouts(listener).get(0).pollingBaseline, nullValue());
+    }
+
+    @Test
+    @Issue("NGPIPELINE-917")
+    // Showing how only with both changelog/poll false is when it removes the checkout
+    public void baselineResetSingleRepoEdgeCases() throws Exception {
```
alexanderrtaylor: This test probably needs to be changed/removed because this is "incorrect behavior"

I will do an ignore on this for now and show the correct behavior

bitwiseman: The "incorrect behavior" here is really only "less than ideal". It is a separate bug that should be described and fixed separately.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -491,4 +498,208 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+        r.assertBuildStatusSuccess(r.waitForCompletion(b4));
+        assertThat(b4.checkouts(listener).size(), equalTo(1));
+        //Proving it is reset
+        assertThat(b4.checkouts(listener).get(0).pollingBaseline, nullValue());
+    }
+
+    @Test
+    @Issue("NGPIPELINE-917")
+    // Basic test to prove that checkout/poll gets individually set for each checkout
+    public void baselineResetMultipleRepos() throws Exception {
```
dwnusbaum: nit: I feel like between the single-repo tests and `baselineResetMultipleSameRepo` we cover all of the interesting behavior, so I'd just delete this test.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -491,4 +498,208 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
         assertEquals(1, checkouts.size());
         assertEquals(GitSCM.class, checkouts.get(0).getClass());
     }
+
+    @Test
+    @Issue("NGPIPELINE-917")
```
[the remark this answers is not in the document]
dwnusbaum: I think this is the only thing I didn't in my PR, but it should be done before this is merged.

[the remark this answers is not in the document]
alexanderrtaylor: OK thanks that is the only place it is reported officially  so I wanted to get something in there. I will remove

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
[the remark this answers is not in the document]
alexanderrtaylor: @dwnusbaum So this is eventually what I would like to start failing because this is not the ideal scenario. Ideally you could actually turn off just changelog and it would turn off both so I kinda wanted a base case and then more nuanced to reveal the difference in behavior.

I am OK with showing both here I just wanted to see if that changed your perspective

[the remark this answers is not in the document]
dwnusbaum: > Ideally you could actually turn off just changelog and it would turn off both so I kinda wanted a base case and then more nuanced to reveal the difference in behavior. 

I don't think we could make that change directly, since it might be a breaking change for some users, so I think the current behavior is here to stay.

Some things I think we could do in workflow-scm-step to help users:

* Log a warning if a users specifies `changelog: true` and `poll: false` explaining that `poll: false` is overriden by `changelog: true`
* Offer an alternate parameter, like `mode` or something, and have 3 options: `CHANGELOG_AND_POLL`, `POLL_ONLY`, `NONE` (bad names but you get the idea), so that users don't have a way to set `changelog: true` and `poll: false`. Probably not worth the complexity since we'd still need to support `changelog` and `poll` in addition to the new option.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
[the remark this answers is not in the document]
dwnusbaum: Better to create an OSS issue, not just remove it here. We are fixing it in OSS, so it's helpful for it to be reported in OSS, that way we can reference the ticket if there are any regressions.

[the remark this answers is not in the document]
dwnusbaum: I created [JENKINS-61415](https://issues.jenkins-ci.org/browse/JENKINS-61415) to track it.

dwnusbaum:
CI build failed due to Windows-specific infra issues, merging anyway.

PR #154, opened 2020-03-18
Use Boolean.equals instead of == when comparing Boolean objects

dwnusbaum:
It is safer to use `Boolean.equals` instead of `==` when comparing nullable `Boolean` objects in case one of the objects was constructed via `new Boolean(boolean)`, which creates distinct objects. 

I don't think that this code ever uses `new Boolean(boolean)`, but it doesn't hurt to make the code more robust just in case.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -421,7 +421,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                     setResult(Result.FAILURE);
                     modified = true;
                 }
-                if (completed != Boolean.TRUE) {
+                if (!Boolean.TRUE.equals(completed)) {
                     completed = true;
                     modified = true;
                 }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -521,23 +521,23 @@ private String key() {
...
+                if (execution != null && !Boolean.TRUE.equals(completed)) {
                     FlowExecution fetchedExecution = getExecution();  // Triggers execution.onLoad so we can resume running if not done
 
                     if (fetchedExecution != null) {
                         if (completed == null) {
                             completed = Boolean.valueOf(fetchedExecution.isComplete());
                         }
 
-                        if (!completed == Boolean.TRUE) {
+                        if (Boolean.FALSE.equals(completed)) {
```
dwnusbaum: Unrelated to the other changes. The old code was fine, since we check `completed == null` a few lines up and set it to a non-null value, but unnecessarily roundabout, since it was unboxing the `Boolean` to a `boolean`, then inverting that and comparing the inverted result to `Boolean.TRUE`, which was also unboxed into a `boolean`. Much clearer in my opinion to just compare to `Boolean.FALSE` directly.

PR #155, opened 2020-03-20
[JENKINS-61604] Do not synchronize on metadataGuard in WorkflowRun.getListener to avoid deadlock

dwnusbaum:
See [JENKINS-61604](https://issues.jenkins-ci.org/browse/JENKINS-61604). Amends #137.

This PR just makes `completed` volatile, which I think results in reasonable behavior. `getListener` might hand out a log that is about to be closed because the build is completing, but it will never open a new log once the build has been marked as closed.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -192,7 +192,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
      * Protects access to {@link #completed} etc.
      * @see #getMetadataGuard
      */
-    private transient Object metadataGuard = new Object();
+    private transient volatile Object metadataGuard = new Object();
 
     /** JENKINS-26761: supposed to always be set but sometimes is not. Access only through {@link #checkouts(TaskListener)}. */
     private @CheckForNull List<SCMCheckout> checkouts;
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -207,9 +207,11 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
      *  Synchronizing this helps ensure that fields are not mutated during a {@link #save()} operation, since that locks on the Run.
      */
     private Object getMetadataGuard() {
-        synchronized (this) {
-            if (metadataGuard == null) {
-                metadataGuard = new Object();
+        if (metadataGuard == null) {
+            synchronized (this) {
+                if (metadataGuard == null) {
+                    metadataGuard = new Object();
+                }
             }
         }
         assert !Thread.holdsLock(this) || Thread.holdsLock(metadataGuard): "Synchronizing on WorkflowRun before metadataGuard may cause deadlocks";
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -186,9 +186,12 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
      * Flag for whether or not the build has completed somehow.
      * This was previously a transient field, so we may need to recompute in {@link #onLoad} based on {@link FlowExecution#isComplete}.
      */
-    Boolean completed;  // Non-private for testing
+    volatile Boolean completed;  // Non-private for testing
 
-    /** Protects access to {@link #completed} etc. */
+    /**
+     * Protects access to {@link #completed} etc.
+     * @see #getMetadataGuard
+     */
     private transient Object metadataGuard = new Object();
 
     /** JENKINS-26761: supposed to always be set but sometimes is not. Access only through {@link #checkouts(TaskListener)}. */
@@ -203,10 +206,13 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
      *  Note: to avoid deadlocks, when nesting locks we ALWAYS need to lock on the guard first, THEN the WorkflowRun.
      *  Synchronizing this helps ensure that fields are not mutated during a {@link #save()} operation, since that locks on the Run.
      */
-    private synchronized Object getMetadataGuard() {
-        if (metadataGuard == null) {
-            metadataGuard = new Object();
+    private Object getMetadataGuard() {
+        synchronized (this) {
+            if (metadataGuard == null) {
+                metadataGuard = new Object();
+            }
         }
+        assert !Thread.holdsLock(this) || Thread.holdsLock(metadataGuard): "Synchronizing on WorkflowRun before metadataGuard may cause deadlocks";
         return metadataGuard;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -203,10 +206,13 @@ public String url() {
...
-    private synchronized Object getMetadataGuard() {
-        if (metadataGuard == null) {
-            metadataGuard = new Object();
+    private Object getMetadataGuard() {
+        synchronized (this) {
+            if (metadataGuard == null) {
+                metadataGuard = new Object();
+            }
         }
+        assert !Thread.holdsLock(this) || Thread.holdsLock(metadataGuard): "Synchronizing on WorkflowRun before metadataGuard may cause deadlocks";
```
dwnusbaum: I added an assertion for the condition given in the method's Javadoc to try to avoid any future issues. It causes a few tests to fail without the changes to `getListener` in this PR, mostly in `CpsPersistenceTest`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -216,13 +222,11 @@ private synchronized Object getMetadataGuard() {
     /** Used internally to ensure listener has been initialized correctly. */
     private BuildListener getListener() {
         // Un-synchronized to prevent deadlocks (combination of run and metadataGuard)
```
dwnusbaum: I guess we didn't notice this in #137, but it's definitely still accurate!

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -579,28 +583,30 @@ private String key() {
     private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
         try {
             setResult(r);
+            BuildListener myListener;
             synchronized (getMetadataGuard()) {
+                myListener = getListener();
```
dwnusbaum: These changes are not strictly necessary, but since I was looking at all uses of `getListener` anyway, I think it's much clearer to call `getListener` once inside of the synchronized block and then use the local variable rather than having this method call `getListener()` repeatedly in addition to accessing the `listener` field directly.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -218,11 +224,9 @@ private BuildListener getListener() {
         // Un-synchronized to prevent deadlocks (combination of run and metadataGuard)
         // Note that in portions where multithreaded access is possible we are already synchronizing on metadataGuard
```
dwnusbaum: I'm not so sure about this comment. There are cases like [this](https://github.com/jenkinsci/workflow-job-plugin/blob/94fa4b91dcf3fbefc4310f97d18ba80eb5372c12/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L461-L463) where the comment looks accurate, but there are also cases like [this](https://github.com/jenkinsci/workflow-job-plugin/blob/94fa4b91dcf3fbefc4310f97d18ba80eb5372c12/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L402) which are clearly happening from various threads with no synchronization, and I don't want to break things by adding synchronization to make the comment accurate.

I think that in practice the listener is initialized very early when the `WorkflowRun` is created, and no other threads are acting on the `WorkflowRun` at that point, and then since `listener` is volatile we never really execute the body of the if statement on the next line until after the build completes, so we don't have to worry about double-initialization.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -186,10 +186,13 @@ public String url() {
...
-    Boolean completed;  // Non-private for testing
+    volatile Boolean completed;  // Non-private for testing
 
-    /** Protects access to {@link #completed} etc. */
-    private transient Object metadataGuard = new Object();
+    /**
+     * Protects access to {@link #completed} etc.
+     * @see #getMetadataGuard
+     */
+    private transient volatile Object metadataGuard = new Object();
```
dwnusbaum: Needs to be volatile now that `getMetadataGuard` does not lock the `WorkflowRun` for its entire body.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -203,10 +206,13 @@ public String url() {
      *  Note: to avoid deadlocks, when nesting locks we ALWAYS need to lock on the guard first, THEN the WorkflowRun.
      *  Synchronizing this helps ensure that fields are not mutated during a {@link #save()} operation, since that locks on the Run.
```
dwnusbaum: I don't totally understand this line. I would like to delete `metadataGuard` completely and just synchronize on `this` instead when modifying metadata, so that we don't have to worry about potential deadlock, but I am concerned that it would cause performance issues.

jglick: > I would like to delete `metadataGuard` completely and just synchronize on `this` instead when modifying metadata, so that we don't have to worry about deadlock

I am tempted to do the same. I would hope that no one is holding a lock on `WorkflowRun` for very long, but not sure.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -203,10 +206,13 @@ public String url() {
...
      */
-    private synchronized Object getMetadataGuard() {
-        if (metadataGuard == null) {
-            metadataGuard = new Object();
+    private Object getMetadataGuard() {
+        synchronized (this) {
+            if (metadataGuard == null) {
+                metadataGuard = new Object();
+            }
         }
```
jglick: Did you mean to use double-checked locking (with `volatile` making it safer)?

```suggestion
        if (metadataGuard == null) {
            synchronized (this) {
                if (metadataGuard == null) {
                    metadataGuard = new Object();
                }
            }
        }
```

dwnusbaum: Oh, good point, I can do that. Really all I wanted to do was be able to add the assertion outside of the scope of the `synchronized` block. Not needing to synchronize on `this` at all when `metadataGuard` is initialized is a nice bonus though.

dwnusbaum: Done in 0c1fde4.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
kshultzCB: I'm pretty far out of my element here, but here goes. Aren't you doing the opposite of this (setting `synchronized (this)` on L211), and then initializing `metadataGuard` after doing so, just above? 

It's also unclear to me when `metaDataGuard` would be `null`, given it was initialized up on L195. This is the first time I've really dug into this so I'm probably just missing something obvious.

dwnusbaum: That's a bit different. Lines 210-216 temporarily lock `this`, and then after that lock is released, the method returns `metadataGuard` for the caller to lock, so both locks are not held simultaneously. The concern here is only if we are holding both locks at the same time, which is what the assertion is checking.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -203,10 +206,15 @@ public String url() {
...
-    private synchronized Object getMetadataGuard() {
+    private Object getMetadataGuard() {
         if (metadataGuard == null) {
-            metadataGuard = new Object();
+            synchronized (this) {
+                if (metadataGuard == null) {
+                    metadataGuard = new Object();
+                }
+            }
         }
```
dwnusbaum: Just for future reference, @olamy reminded me that I was confused why we don't just initialize `metadataGuard` in `readResolve` so that we don't need any locking in this method. Would be worth looking into in a followup PR.

PR #156, opened 2020-03-25
Remove debug text from console.jelly

dwnusbaum:
I added this text to #150 while testing, but forgot to remove it before filing the PR.


dwnusbaum:
I think the CI build is stuck, triggering a new build.

dwnusbaum:
CI infra seems to be having issues. The build passed on some of the branches, and is hanging in other branches (it looks like the build is waiting on an agent). Given the nature of this change, I am going to go ahead and merge it.

PR #159, opened 2020-04-09
Seeing what all fails if WorkflowRun.finish is delayed

jglick:
For investigating #158 and friends.

PR #160, opened 2020-04-09
WorkflowRunRestartTest.resumeDisabled should not assume that the execution is not loaded

jglick:
From #159 I see an apparent mistake in #93. The execution is indeed loaded after a restart:

```
java.lang.Exception: Stack trace
	at java.lang.Thread.dumpStack(Thread.java:1336)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.getExecution(WorkflowRun.java:711)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun$Owner.get(WorkflowRun.java:957)
	at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$1.computeNext(FlowExecutionList.java:65)
	at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$1.computeNext(FlowExecutionList.java:57)
	at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:143)
	at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:138)
	at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$ItemListenerImpl.onLoaded(FlowExecutionList.java:178)
	at jenkins.model.Jenkins.<init>(Jenkins.java:989)
	at …
```

which does not seem improper. The test may have passed accidentally due to a race condition with the restart.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
@@ -99,7 +99,6 @@ public class WorkflowRunRestartTest {
             assertTrue(p.isResumeBlocked());
             WorkflowRun b = p.getBuildByNumber(1);
             r.waitForCompletion(b);
-            assertFalse(b.executionLoaded);
             assertTrue(b.completed);
             assertFalse(b.isBuilding());
             assertEquals(Result.ABORTED, b.getResult());


src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
```diff
@@ -99,7 +99,6 @@
             assertTrue(p.isResumeBlocked());
             WorkflowRun b = p.getBuildByNumber(1);
             r.waitForCompletion(b);
-            assertFalse(b.executionLoaded);
```
dwnusbaum, line 102: Should we invert the assertion instead of deleting it? Would pulling in https://github.com/jenkinsci/jenkins-test-harness/pull/210 make the behavior consistent?

jglick: > Should we invert the assertion

Not sure I see the point.

> Would pulling in jenkinsci/jenkins-test-harness#210 make the behavior consistent?

Not sure.

PR #126, opened 2019-03-07
Pick up JTH changes to skip MemoryAssert in tests on JDK9+

dwnusbaum:
Subsumes #125 and #122 (needed for the workflow-support update).

PR #163, opened 2020-05-04
Fix build errors by updating minimum core version, parent POM, and dependencies

dwnusbaum:
Should fix enforcer errors seen in the CI builds when testing against 2.222.x.

2.176.x is the fourth-oldest LTS line at this point, so it seems like a conservative enough upgrade. The core BOM used by version 4.x of the parent POM also supports 2.164.3 if we want to be even more conservative.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(useAci: true, configurations: buildPlugin.recommendedConfigurations())
+buildPlugin(useAci: false, configurations: buildPlugin.recommendedConfigurations())
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,8 +1,1 @@
-coreJdk11Version="2.161"
-
-buildPlugin(configurations: [
-  [ platform: "linux", jdk: "8", jenkins: null ],
-  [ platform: "windows", jdk: "8", jenkins: null ],
-  [ platform: "linux", jdk: "11", jenkins: coreJdk11Version, javaLevel: "8" ],
-  [ platform: "windows", jdk: "11", jenkins: coreJdk11Version, javaLevel: "8" ]
-])
+buildPlugin(useAci: true, configurations: buildPlugin.recommendedConfigurations())
--- a/pom.xml
+++ b/pom.xml
@@ -68,55 +68,54 @@
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
-        <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
-        <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
-        <git-plugin.version>3.2.0</git-plugin.version>
     </properties>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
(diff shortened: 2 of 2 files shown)


car-roll:
I have yet to see a theme among all these CI failures across repos when doing these upgrades. It seems like each repo has its own unique set of failures. I guess that makes sense since they each have their own dependencies...

dwnusbaum:
One test failure (`WorkflowJobTest.polling`) seems infra related (maybe `mingit` is broken on Windows ACI agents?):

> C:\mingit\usr\bin\sh.exe: *** fatal error - unable to load netapi32.dll, Win32 error 126

The other two (`DefaultLogStorageTest.consoleNotes` and `NewNodeConsoleNoteTest.labels`) look like they are probably caused by some security fix or hardening in the newer version of Jenkins and the tests need to be updated.

dwnusbaum:
Ok, some of the errors were because Git is broken on the Windows ACI agents, and the others were related to Trilead being split from new versions of Jenkins core. I addressed the first issue by switching back to non-ACI agents for now, and the second by using the plugin BOM to update dependencies to pull in a fix from workflow-api that removed an unnecessary Trilead dependency.

pom.xml
```diff
@@ -129,93 +128,53 @@
...
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>timestamper</artifactId>
-            <version>1.9</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.jenkins-ci</groupId>
-            <artifactId>version-number</artifactId>
```
dwnusbaum, line 217: This library was not actually used in tests as far as I can tell. I think it was added to fix an enforcer issue in #126 that is now irrelevant because of the core BOM.

pom.xml
```diff
@@ -129,93 +128,53 @@
...
-            <version>${scm-api-plugin.version}</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
-            <version>${git-plugin.version}</version>
-            <scope>test</scope>
-            <exclusions>
```
dwnusbaum, line 165: It looks like the POM issues that made the exclusions necessary have been fixed in recent versions of these plugins.

Jenkinsfile
```diff
@@ -1,8 +1 @@
-coreJdk11Version="2.161"
-
-buildPlugin(configurations: [
-  [ platform: "linux", jdk: "8", jenkins: null ],
-  [ platform: "windows", jdk: "8", jenkins: null ],
-  [ platform: "linux", jdk: "11", jenkins: coreJdk11Version, javaLevel: "8" ],
-  [ platform: "windows", jdk: "11", jenkins: coreJdk11Version, javaLevel: "8" ]
-])
+buildPlugin(useAci: false, configurations: buildPlugin.recommendedConfigurations())
```
basil, line 1: `useAci` defaults to `false`, so you could just as easily omit it. But of course this is fine as-is, too.

PR #164, opened 2020-05-26
Fix issue with big logs, causing the browser to freeze (JENKINS-56377)

zhekoff316:
See [JENKINS-56377](https://issues.jenkins-ci.org/browse/JENKINS-56377).

We've had this issue for a couple of months, where big logs would freeze the browser and the whole thing would load for ~2-3 minutes. I've addressed this issue and optimized the code in several places:

- Moved "hide/show" link generation functionality in the back-end
- Removed an unnecessary loop, which would loop over every element, while it was looping through each element. This was the main cause of the performance issues.
- I've added the setTimeout() function, so it prevents browser freezing in especially big logs
- I've also added some documentation, so it will be easier in the future for everyone to work on this script and improve it

I've tested it on a pipeline, which has 5-10 thousand steps, and my load time has been in the span of 2-5 seconds, compared to 90-150s before (including a browser crash/freeze). Also, this has been tested on Chrome and Mozilla.


zhekoff316:
I feel like there is an issue with the build process, as a higher version of Jenkins is required to run the tests (the POM states 2.176.4 as the minimum Jenkins version, while the tests ran at 2.164.1). I've ran the tests locally and they all pass,

dwnusbaum:
Thanks for the PR!

> I feel like there is an issue with the build process

Yeah there is, I just filed https://github.com/jenkinsci/workflow-job-plugin/pull/165 to hopefully fix the issue.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,62 +1,79 @@
...
-    if (label != null) {
-        var html = e.innerHTML
-        var suffix = ' (' + label.escapeHTML() + ')';
-        if (html.indexOf(suffix)<0) {
-            e.innerHTML = e.innerHTML.replace(/.+/, '$&' + suffix) // insert before EOL
+/**
+ * The setTimeout() method calls a function or evaluates an expression after a specified number of milliseconds.
+ * It is used to prevent a browser freeze that might occur if a very big log is being evaluated.
+ */
+setTimeout(() => {
```
jglick, line 5: This smells wrong. I would expect the timeout to be applied to the callback, not to the call to `Behaviour.specify` itself.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,62 +1,79 @@
...
+    /**
+     * Specifies something to do when an element matching a CSS selector is encountered.
+     *
+     * @param {String} selector a CSS selector triggering your behavior
+     * @param {String} id combined with selector, uniquely identifies this behavior; prevents duplicate registrations
+     * @param {Number} priority relative position of this behavior in case multiple apply to a given element; lower numbers applied first (sorted by id then selector in case of tie); choose 0 if you do not care
+     * @param {Function} behavior callback function taking one parameter, a (DOM) {@link Element}, and returning void
+     */
```
jglick, line 13: This is all part of the API documentation; should not duplicated at a call site.

```suggestion
```

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,62 +1,79 @@
...
-        enclosings.set(oid, node.getAttribute('enclosingId'))
-        labels.set(oid, node.getAttribute('label'))
-    }
-    var id = nodeId
-    while (true) {
-        id = enclosings.get(id)
-        if (id == null) {
-            break
+
+        // The CSS rule for branch names only needs to be added once per node, so we
```
jglick, line 35: BTW https://github.com/jenkinsci/workflow-job-plugin/pull/164/files?diff=unified&w=1 improves reviewability due to indentation changes.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,62 +1,79 @@
...
+
+        if (e.processedNewNodeConsoleNote) {
+            return
         }
-    }
-    var nodeId = e.getAttribute('nodeId')
-    var startId = e.getAttribute('startId')
-    if (startId == null || startId == nodeId) {
-        e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" onclick="showHidePipelineSection(this); return false">hide</a>)</span>')
-        // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
```
jglick, line 18: I think this is already filed in JIRA somewhere.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
zhekoff316: I do agree with that. Will change it.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
zhekoff316: I removed this part, because it would probably make more sense to generate this inside the back-end.

jglick: I should have clarified that my review comment was in reference to the removal of the one-line code comment, which is properly an RFE that should be tracked in Jira, not to behavioral changes in your patch.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,62 +1,78 @@
...
-    }
-    var nodes = $$('.pipeline-new-node')
-    var enclosings = new Map() // id → enclosingId
-    var labels = new Map() // id → label
-    for (var i = 0; i < nodes.length; i++) {
-        var node = nodes[i]
-        var oid = node.getAttribute('nodeId')
-        enclosings.set(oid, node.getAttribute('enclosingId'))
-        labels.set(oid, node.getAttribute('label'))
-    }
```
dwnusbaum, line 38: Removing this loop causes branch names to no longer be shown when log output changes from one branch to another in many cases.

The problem is that in the `while` loop right after this, we need to walk through all of this node's parents to find what parallel branch it is a part of. Because the loop is gone, now we only walk through the immediate parents of each node, so if there is any nesting inside of the parallel branch, for example because of a block-scope step, we never find the enclosing parallel branch, and so the `while` loop exits before finding a label.

The old code is not optimal because we create a fresh `enclosings` and `labels` map for every node even though most of the content will be the same. I would try to come up with a good way to store those as page-level state so that you can keep the approach you have now of only updating `enclosings` and `labels` once per node, and then split the `while` loop below out somehow and only run it after the `enclosings` and `labels` have data for every node (not sure of a good way to do that).

Here is a Pipeline you can run to compare the behavior with and without this PR to see what I mean:

```
stage('Outer') {
    parallel(branchOne: {
        stage('Stage One') {
            node() {
                sh 'echo "from one"'
                sleep(time: 1)
                echo 'from one again'
            }
        }
    }, branchTwo: {
        stage('Stage Two') {
            node() {
                sh 'echo "from two"'
                sleep(time: 1)
                echo 'from two again'
            }
        }
    })
}
```

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,62 +1,78 @@
...
+            var label = labels.get(id)
+            if (label != null && label.indexOf('Branch: ') == 0) {
+                var branch = label.substring(8)
+                var ss = document.styleSheets[0]
+                // TODO https://stackoverflow.com/a/18990994/12916 does not scale well to add width: 25em; text-align: right
+                ss.insertRule('.pipeline-node-' + nodeId + '::before {content: "[' + branch.escapeHTML() + '] "; color: #9A9999; position: absolute; transform: translateX(-100%)}', ss.cssRules == null ? 0 : ss.cssRules.length)
+                break
+            }
+        }
+    }, 5000)
```
dwnusbaum, line 62: 5 seconds is a long time to wait. Is 1 second (or maybe even less) not good enough just to get it out of the critical path for the main page load?

dwnusbaum (dismissed):
@zhekoff316 Thanks for the PR! I am requesting changes to fix the issue I noted with parallel branch labels, but I would really like to see this merged once that is fixed!

> I've tested it on a pipeline, which has 5-10 thousand steps

Is that a real Pipeline, or just one you created for testing?

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
zhekoff316: I see. I will try to find a solution to this problem.

zhekoff316:
Hello again, thanks for the feedback.

First of all, sorry about the delay.

I have done some big changes in the latest commit. They resolve the issue with parallel pipelines (branch names are shown), and also greatly improve the performance with big logs.

- I have removed the usage of the Behaviour.js library, as we can use basic JavaScript to accomplish the same results with greater performance. Also, this library is deprecated.
- I have split most of the logic into separate functions. In my opinion, this enhances readability and maintainability.
- More technically speaking, the problematic for loop, which generates the enclosings now only executes once, and if the pipeline is parallel it will go through another function, which will append the branch names. I feel like this is the most optimal way to do it while preserving all current functionality.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,41 +1,52 @@
...
-    // a duplicate synthetic span element for this node.
-    var maybeDupeNodes = $$('[nodeid=\"'+nodeId+'\"].pipeline-new-node');
-    for (var i = 0; i < maybeDupeNodes.length; i++) {
-        var node = maybeDupeNodes[i];
-        if (node !== e && node.processedNewNodeConsoleNote) {
-            return;
-        }
+var isParallel = false
+var nodes = $$('.pipeline-new-node')
+var pipelineElements = document.getElementsByClassName("pipeline-new-node")
```
dwnusbaum, line 3: Is this used anywhere?

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -127,3 +189,8 @@ function showHidePipelineSection(link) {
         }
     }
 }
+
+window.addEventListener('load', function() {
+    isParallelBuild($$('.pipeline-new-node'));
+    generatePipelineLinks($$('.pipeline-new-node'));
```
dwnusbaum, line 195: Doing this only once on page load seems like it will have issues if someone opens the log for an in-progress build, leaves it open while the build continues, and then comes back and looks at the logs. For example, there may not be a parallel step when the page is opened initially, but one may execute later on. I think that in-progress builds are supported today by `Behaviour.specify` and the old code, but I haven't specifically tested it.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
zhekoff316: Correct, those two variables are left-over after I made some changes to the code. Will remove them.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
zhekoff316: I will test this use case with the current code and will let you know. It is also a variant to remove the parallel check all-together - the performance impact will be minimal, as the main performance issue was fixed.

oleg-nenashev:
Added it to my review queue. Thanks to @zhekoff316 for this pull request, it is a very important improvement!

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,41 +1,62 @@
...
+    enclosings.set(oid, node.getAttribute('enclosingId'))
+    labels.set(oid, node.getAttribute('label'))
+}
+
+/**
+ * Generate enclosings and labels for all of the pipeline elements
+ */
+function generateEnclosings () {
     var enclosings = new Map() // id → enclosingId
     var labels = new Map() // id → label
```
dwnusbaum, line 43: Did you mean to create a local version of `enclosings` and `labels` here, or were you wanting to reuse and modify the global variables and use them as a kind of cache or something?

The way things are now, it doesn't look like `generateEnclosing` actually does anything because the global variables aren't used.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,41 +1,62 @@
-Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e) {
-    if (e.processedNewNodeConsoleNote) {
-        return
-    }
-    e.processedNewNodeConsoleNote = true
-    var label = e.getAttribute('label')
+var enclosings = new Map() // id → enclosingId
+var labels = new Map() // id → label
```
dwnusbaum, line 2: I'm not sure if scripts like this are encapsulated by Jenkins beyond standard JavaScript behavior, but having global vars like this seems like a bad idea. I would probably find the `<pre>` element with class `console-output` and attach these objects to that element instead, but I'm not a JavaScript expert, so I'm not sure what the preferred approach is for this kind of thing.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,41 +1,62 @@
...
     var labels = new Map() // id → label
+    var nodes = $$(".pipeline-new-node")
+
     for (var i = 0; i < nodes.length; i++) {
         var node = nodes[i]
         var oid = node.getAttribute('nodeId')
+
         enclosings.set(oid, node.getAttribute('enclosingId'))
         labels.set(oid, node.getAttribute('label'))
+        appendBranchNames(oid, enclosings, labels)
```
dwnusbaum, line 52: It looks like this is the only place where `appendBranchNames` is called, which seems wrong, since this method is only called for the `parallel` step itself. What if we start executing the first branch in a `parallel` step, and then that branch stops and we start executing the other branch? The first branch might get labelled correctly, but I think that the second branch wouldn't be labeled (unless later on in the Pipeline there was another `parallel` step).

ngg:
Hi! Is there any update regarding this PR? We hit this bug as well :(

bitwiseman:
@ngg 
We're waiting for updates from @zhekoff316, I believe.

zhekoff316:
I've been on a vacation, but I will take a look at the latest review comments as soon as I can, and will let you know if I can fix it.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
ngg: I think it'd be easier to simply wrap the whole script in an IIFE
```js
(function() {
    // ...
})();
```
This way those variables are not in the global scope but easily accessible from anywhere in this script.

zhekoff316:
Hello,

I've tried to address the latest comments from before I stopped working on this PR.

(1) The global variables were indeed useless there, so I removed them and replaced them with variables in the function scope
(2) Regarding your comments  @dwnusbaum  about the `appendBranchNames`:

> It looks like this is the only place where appendBranchNames is called, which seems wrong, since this method is only called for the parallel step itself. What if we start executing the first branch in a parallel step, and then that branch stops and we start executing the other branch? The first branch might get labelled correctly, but I think that the second branch wouldn't be labeled (unless later on in the Pipeline there was another parallel step).

Could you please give me an example pipeline where I can test this for example, because I can not reproduce anything wrong currently with the parallel builds. This function is only called if a parallel stage exists in the whole pipeline, and will go through all elements (from what I can see). Maybe I am doing something wrong, so it would be very helpful if you can provide me a very short example :) If there is something wrong there, I will fix it.

-----

Overall, I tested again with a job with a lot of Pipeline Steps  and the log would open very fast. Parallel builds with multiple stages also looked okay, and the show/hide buttons are working. I am ready to provide examples if you need them.

Will be happy if you can review this, as it's quite an important improvement.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -1,41 +1,62 @@
...
+    labels.set(oid, node.getAttribute('label'))
+}
+
+/**
+ * Generate enclosings and labels for all of the pipeline elements
+ */
+function generateEnclosings () {
     var enclosings = new Map() // id → enclosingId
     var labels = new Map() // id → label
+    var nodes = $$(".pipeline-new-node")
```
halkeye, line 44: in general, document.querySelectorAll will be much faster than $$ as one is a native browser API, and the other would be a prototypes scanner. its not a drop in replacement, but it doesn't seem like there's any prototype specific apis being used for this, so should work.

jessejoe:
I've also been hitting this issue as part of testing upgrading our old Jenkins server (`v2.150.1`) to current LTS, but this issue makes `/console` and `/consoleFull` pretty much unusable for us (I've had to use `workflow-job-plugin` `v.2.36` or earlier so far). I've documented a lot of my troubleshooting in https://issues.jenkins.io/browse/JENKINS-62241.

I just tested the changes in this PR with a build log from an example pipeline in the ticket I mentioned above which is simply generating a log from parallel branches:
```groovy
def genText(lines){
    (1..lines).each{
       println "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur interdum fringilla interdum"
    }
    return true
}
parallel firstBranch: {
    genText(20000)
}, secondBranch: {
    genText(20000)
}
```

I wanted to give some benchmarks of trying to load the different consoles for the the result of a build of the above pipeline, which winds up being a ~16MB `log` file:

| Jenkins Version | `workflow-job-plugin` Version | `/console` | `/consoleFull`  | `/consoleText` |
|  -------------  | ------------- | ------------- | ------------- | ------------- |
| `2.150.1` (our existing server) | `2.28`  | <1 second | 5-7 seconds | <1 second |
| `2.319.2` (local container) | `2.36` | <1 second | Never (5 min timeout) - browser slowdown warning |  <1 second |
| `2.319.2` (local container) | `1145.v7f2433caa07f`  | Never (5 min timeout) - browser slowdown warning | Never (5 min timeout) - browser slowdown warning |  <1 second |
| `2.319.2` (local container) | `zhekoff316:master:43c58` | 5-7 seconds | 65-75 seconds | <1 second |

The changes in this PR are definitely an improvement over the more current versions, but still not as efficient as our old Jenkins server `2.150.1` with `workflow-job-plugin` `2.28`. Definitely at least make it usable though!

dwnusbaum:
The difference between  1145.v7f2433caa07f and 2.36 is that prior to #150 (landed in 2.37), the relevant Javascript essentially never executes on `/console` for large log files due to the bug described in that PR, only on `/consoleFull` (this is why they have the same behavior on `/consoleFull`). The difference between 2.36 and 2.28 is that that https://github.com/jenkinsci/workflow-job-plugin/pull/111 (landed in 2.31) added Javascript to try to show the names of parallel branches in the console UI, and this Javascript is slow because it traverses the Pipeline flow graph for each flow node that produced log output to figure out what branch each node belongs to (IIRC this PR (#164) tries to improve the performance by caching the intermediate steps of the traversal). Prior to https://github.com/jenkinsci/workflow-job-plugin/pull/27 (landed in 2.26) the branch names were written directly into the logs, so there was no Javascript needed to show this data at all.

So in 2.28, there is no Javascript that tries to show branch names in the log, which is why things are fast.

jglick:
Possibly we could just detect a large log file and disable the JS (compared to `consoleText` you would still see rendered console notes)? Probably needs bigger redesigns to make the display usable on a large build, especially when there are parallel branches involved.

jessejoe:
After spending the week discussing this in the Jenkins Gitter, it seems like most large log files do not hit this issue (or I suspect there's be a lot more complaints). The magic ingredient seems to be the multiple parallel branches logging (could detect that instead?). IMO the biggest issue is that even `/console` is affected, which really doesn't even show a lot of the log, and it is linked to from the job build history and build executor status. As a user, I think `/console` should load as fast as possible since its such a common endpoint and should not take multiple seconds to load a little bit of log.

wblair-ebsco:
I hit this problem with large console logs quite often with version 2.293. Haven't complained but that does not means it's not a problem.
Behaves like a memory leak because I have to restart the controller to display large console logs.

PR #165, opened 2020-05-26
Replace use of recommendedPlugins with explicit configurations in Jenkinsfile

dwnusbaum:
#163 fixed the build after it was broken by https://github.com/jenkins-infra/pipeline-library/pull/92, but then https://github.com/jenkins-infra/pipeline-library/pull/146 broke it again.

This PR stops using `buildPlugin.recommendedConfigurations` to avoid these kinds of issues in the first place.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,5 @@
-buildPlugin(useAci: false, configurations: buildPlugin.recommendedConfigurations())
+buildPlugin(useAci: false, configurations: [
+  [ platform: "linux", jdk: "8", jenkins: null ],
+  [ platform: "windows", jdk: "8", jenkins: null ],
+  [ platform: "linux", jdk: "11", jenkins: "2.222.3", javaLevel: 8 ]
+])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 buildPlugin(useAci: false, configurations: [
-  [ platform: "linux", jdk: "8", jenkins: null ],
-  [ platform: "windows", jdk: "8", jenkins: null ],
-  [ platform: "linux", jdk: "11", jenkins: "2.222.3", javaLevel: 8 ]
+  [ platform: "linux", jdk: "8" ],
+  [ platform: "windows", jdk: "8" ],
+  [ platform: "linux", jdk: "11", jenkins: "2.222.3" ]
 ])


Jenkinsfile
```diff
@@ -1 +1,5 @@
-buildPlugin(useAci: false, configurations: buildPlugin.recommendedConfigurations())
+buildPlugin(useAci: false, configurations: [
+  [ platform: "linux", jdk: "8", jenkins: null ],
+  [ platform: "windows", jdk: "8", jenkins: null ],
+  [ platform: "linux", jdk: "11", jenkins: "2.222.3", javaLevel: 8 ]
```
kshultzCB, line 4: There might be some context I've missed in the other PRs that led up to this one, but I'll ask anyway. Why specify a Jenkins version here, but not in the other two?

dwnusbaum: `jenkins: null` means to use the version specified in `pom.xml` under `<jenkins.version>`. We could explicitly specify `2.176.4` here instead, but then anytime we updated the version we would need to update both `pom.xml` and `Jenkinsfile`, which is easy to forget (not really a problem since the PR build will fail until you fix it, it's just a bit tedious). This way, we only need to update `Jenkinsfile` if we want to test on a newer version than `2.222.3`.

dwnusbaum: Oh and the big idea here is to test on Windows and Linux using Java 8 and the minimum required Jenkins version, then on Linux using Java 11 and a much newer LTS to make sure we don't break things for the new LTS or for Java 11.

Jenkinsfile
```diff
@@ -1 +1,5 @@
-buildPlugin(useAci: false, configurations: buildPlugin.recommendedConfigurations())
+buildPlugin(useAci: false, configurations: [
```
jglick, line 1:
```suggestion
buildPlugin(configurations: [
```

as this is the default

Jenkinsfile
```diff
@@ -1 +1,5 @@
...
+  [ platform: "linux", jdk: "8", jenkins: null ],
+  [ platform: "windows", jdk: "8", jenkins: null ],
```
jglick, line 3: KISS

```suggestion
  [ platform: "linux", jdk: "8" ],
  [ platform: "windows", jdk: "8" ],
```

Jenkinsfile
jglick, line 4:
```suggestion
  [ platform: "linux", jdk: "11", jenkins: "2.222.3" ]
```

Jenkinsfile
dwnusbaum: I want this to be explicit in case the default changes in the future since the version of Git installed on the Windows ACI agents currently breaks this plugin's tests.

dwnusbaum:
Rebuilding:

> [2020-05-28T20:57:15.881Z] EC2 (aws) - Ubuntu 18.04 LTS (i-0fe3f98bb2db782d4) was marked offline: Node is being removed
> Unable to create live FilePath for EC2 (aws) - Ubuntu 18.04 LTS (i-0fe3f98bb2db782d4)

dwnusbaum:
> EC2 (aws) - Ubuntu 18.04 LTS (i-0ea714d759e16f365) was marked offline: Connection was broken: java.io.EOFException

PR #172, opened 2020-10-13
Add node indicator to build overview page

lgoenner:
Pipeline builds currently do not show the node they were built on in the build overview page:
image

In case only a single node is used during a pipeline build, it should be displayed the same way as in a freestyle build:
image

Reading nodes from WorkspaceActions seemed the most sensible to me, as it works for old builds that have already been completed and requires no modifications to other pipeline plugins. The jelly implementation is very similar to [Jenkins core](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/model/AbstractBuild/index.jelly#L44), so it does not need additional localization or components. This does limit the feature to builds that only use a single node, but I would guess that that is the most common case. 
Perhaps a text saying 'on multiple nodes' might be an alternative for builds that used multiple nodes.

I'd be happy to hear your feedback on this! Since I'm contributing this as part of Hacktoberfest, I'd appreciate it if you could tag this PR with hacktoberfest-accepted as well.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.workflow.job;
 
 import com.google.common.collect.ImmutableSortedSet;
+import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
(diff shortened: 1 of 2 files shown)


dwnusbaum (changes requested):
Thanks for the PR! I'm not sure if making a change like this that only applies to some Pipelines is something that we should do. Also, I think this kind of change is related to [JENKINS-38877](https://issues.jenkins-ci.org/browse/JENKINS-38877).

From my point of view, the "workspaces" link in the sidebar for Pipeline builds is the right way to show this kind of information should be shown for Pipelines. The fact that it is on its own page is a good thing, because it means that there is no need to load every flow node for a build unless and until someone actually wants to know what agents it unused. If the information on that page is not clear, or you think the format could be improved, I would be happy to consider changes to it (see `WorkspaceRunAction` in https://github.com/jenkinsci/workflow-support-plugin).

What do you think?

jglick:
Seems to have been abandoned, and it is not obvious whether the proposed change could be implemented without a performance penalty.

PR #184, opened 2021-03-16
Ensure pipeline error messages are printed before RunListener.fireCompleted()

GLundh:
By not writing stack traces to the console log before executing
RunListener.fireCompleted() we cannot carry out error analysis
from upstream jobs before the downstream builds are released.

This brings issues with other Extensions like the Build Failure Analyzer.

It's reasonable that any important error messages concerning the flow
execution is printed before the Run is completed.

Example:
https://issues.jenkins.io/browse/JENKINS-42755

In the linked issue, the fix was erroneously "fixed" in BFA, with bad
side effects, like downstream builds being BFA-analyzed *after* the
waiting upstream build is completed and the possibility to correctly
aggregate BFA findings in upstream jobs are gone.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -607,8 +607,6 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                 // Never even made it to running, either failed when fresh-started or resumed -- otherwise getListener would have run
                 LOGGER.log(Level.WARNING, this + " failed to start", t);
             } else {
-                RunListener.fireCompleted(WorkflowRun.this, myListener);
-                fireCompleted();
                 if (t instanceof AbortException) {
                     myListener.error(t.getMessage());
                 } else if (t instanceof FlowInterruptedException) {
@@ -616,6 +614,8 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                 } else if (t != null) {
                     Functions.printStackTrace(t, myListener.getLogger());
                 }
+                RunListener.fireCompleted(WorkflowRun.this, myListener);
+                fireCompleted();
                 myListener.finished(getResult());
                 if (myListener instanceof AutoCloseable) {
                     try {


GLundh:
@dwnusbaum Would you mind taking a look. It is a small but important (to us) change.

dwnusbaum (commented):
@GLundh Thanks for the PR! My guess is that the code is the way it is so that any messages added to the build log by any `RunListener` extensions always come before the ultimate cause of the build failing for clarity. I see how this is not optimal for a plugin like https://plugins.jenkins.io/build-failure-analyzer/ but I am a little concerned that there might be some confusion/problems if we just move this code around.

Can BFA use `RunListener.onFinalized` (called after the build log is completely closed) instead of `RunListener.onCompleted`? Or does it need to be able to write to the build log?

Also, looking at the history of this code, I don't see any particular reason why we couldn't change the ordering in the code, so my above concern is not necessarily a blocker, just something that I would consider. The only somewhat related change I saw was https://github.com/jenkinsci/workflow-job-plugin/pull/143.

GLundh:
> Can BFA use `RunListener.onFinalized` (called after the build log is completely closed) instead of `RunListener.onCompleted`? Or does it need to be able to write to the build log?
> 
> Also, looking at the history of this code, I don't see any particular reason why we couldn't change the ordering in the code, so my above concern is not necessarily a blocker, just something that I would consider. The only somewhat related change I saw was #143.

Thanks for getting back to me, @dwnusbaum !
Ok. So I have done a pretty through analysis of the issue. Sorry if some information is redundant, I'm copying + pasting most of this text from an internal issue.

1. `RunListener.fireCompleted()` is executed **before** a build is _released_. When we say _released_ we mean that any waiting upstream builds can continue (or downstream builds can be triggered).

2. `RunListener.fireFinalized()` is executed **after** has build been _released_.

The Build Failure Analyzer plugin was actually written to be executing in `RunListener.onCompleted()` scope and then **wait** for the analysis to be completed. This means that the waiting upstream build will wait for its downstream builds to complete (and be analysed) before doing its own aggregation and analysis.

See the `waitForAllTasks()` method in BFA, which originally was executed *only* within the `RunListener.onCompleted()` scope: https://github.com/jenkinsci/build-failure-analyzer-plugin/blob/master/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java#L497

Typically execution flow:
```
BuildFailureScanner.waitAllTasks(PrintStream, List<Future<?>>)  (com.sonyericsson.jenkins.plugins.bfa)
BuildFailureScanner.findIndications(Collection<FailureCause>, Run, PrintStream)  (com.sonyericsson.jenkins.plugins.bfa)
BuildFailureScanner.findCauses(Collection<FailureCause>, Run, PrintStream)  (com.sonyericsson.jenkins.plugins.bfa)
BuildFailureScanner.scan(Run, PrintStream)  (com.sonyericsson.jenkins.plugins.bfa)
BuildFailureScanner.scanIfNotScanned(Run, PrintStream)  (com.sonyericsson.jenkins.plugins.bfa)
BuildFailureScanner.doScan(Run)  (com.sonyericsson.jenkins.plugins.bfa)
BuildFailureScanner.onCompleted(Run, TaskListener)  (com.sonyericsson.jenkins.plugins.bfa)
```

However, due to the implementation of printing pipeline errors in `WorkflowRun`, the actual pipeline error messages could not be picked up by BFA from `RunListener.fireCompleted()`. (Hence this PR)

This meant that BFA could not react on pipeline error messages until someone posted this PR to BFA: https://github.com/jenkinsci/build-failure-analyzer-plugin/pull/103

This PR essentially fixes so that BFA is executed in `RunListener.onFinalized()` (liked you asked about) for non `AbstractBuilds`, so that the `WorkflowRun` pipeline error messages can be picked up and parsed. *This is problematic, since the upstream job is now already given the OK to continue executing _before_ the BFA analysis is executed.* And we will see a really bad racing condition where the upstream build could complete its log analysis before the downsteam build. This is extra problematic, if you want to emit messages (mq-notifier plugin) with this kind of data aggregated.

I argue that the correct fix should have been in the WorkflowRun, ensuring that any error messages are printed in the log before the `RunListener.fireCompleted()` is executed. AFAIK, this is how most job types handle it.

If this PR is accepted, it should be enough to once again allow BFA to analyze in the RunListener.onCompleted() context, and we can solve issues related to race conditions and aggregations of log errors.

bitwiseman:
> My guess is that the code is the way it is so that any messages added to the build log by any `RunListener` extensions always come before the ultimate cause of the build failing for clarity.

This seems exactly right. If we move these, various "completed" actions could log info after cause.  

I'm not saying this is a good idea, but what about adding the failure cause to the `Completed` call? 

Or maybe it is okay for completed events to  add messages  after the raw error?

dwnusbaum:
> [...] any error messages are printed in the log before the RunListener.fireCompleted() is executed. AFAIK, this is how most job types handle it.

If this is true (I'm not sure, but from a quick glance at `Run.execute` it does look like the [final build failure message is printed to logs](https://github.com/jenkinsci/jenkins/blob/7c0d4df8cb6c2b5621b8e71daee6773b1e47715d/core/src/main/java/hudson/model/Run.java#L1912-L1928) before the call to [`RunListener.onFinalized`](https://github.com/jenkinsci/jenkins/blob/7c0d4df8cb6c2b5621b8e71daee6773b1e47715d/core/src/main/java/hudson/model/Run.java#L1951)), then it is probably fine to make Pipelines match the common behavior. Maybe worth running a PCT test against https://github.com/jenkinsci/bom just to see if anything comes up.

GLundh:
> If this is true (I'm not sure, but from a quick glance at `Run.execute` it does look like the [final build failure message is printed to logs](https://github.com/jenkinsci/jenkins/blob/7c0d4df8cb6c2b5621b8e71daee6773b1e47715d/core/src/main/java/hudson/model/Run.java#L1912-L1928) before the call to [`RunListener.onFinalized`](https://github.com/jenkinsci/jenkins/blob/7c0d4df8cb6c2b5621b8e71daee6773b1e47715d/core/src/main/java/hudson/model/Run.java#L1951)), then it is probably fine to make Pipelines match the common behavior. Maybe worth running a PCT test against https://github.com/jenkinsci/bom just to see if anything comes up.

"before the call to [`RunListener.onFinalized`]" <- I think you actually meant [RunListener.onCompleted()](https://github.com/jenkinsci/jenkins/blob/7c0d4df8cb6c2b5621b8e71daee6773b1e47715d/core/src/main/java/hudson/model/Run.java#L1951) here.

GLundh:
Also MavenJob prints errors before the [`RunListener.fireCompleted()`](https://github.com/jenkinsci/maven-plugin/blob/bf8facf6bee34ef1a62a0ad8fa584551fa397ce8/src/main/java/hudson/maven/MavenBuild.java#L617-L636).

In the end, I really see no decent way to solve my issue, except fixing it in the `WorkflowRun` (where it should be fixed IMHO). I cannot move the analysis to `RunListener.onFinalized` due to the reasons previously iterated.

Is there any way I can move this along? Someone mentioned the Jenkins Bill of Materials. I have really no idea what you want me to do here. Can I use that to see if I break other plugins, without issuing a new release of WorkflowJob?

GLundh:
> This seems exactly right. If we move these, various "completed" actions could log info after cause.
> 

Yes. This seems to be the common way of doing it in other job types.

I would argue that it is far more important that we give  plugins a way to analyse issues before allowing upstream jobs to continue (or triggering downstream jobs) than avoiding a few extra lines after the error cause.

But then I'm a bit frustrated by this bug :)

bitwiseman (approved):
Sounds good.

GLundh:
@car-roll are you OK with this too?

car-roll:
I put the incremental up in the bom latest here: https://github.com/jenkinsci/bom/pull/468

timja:
> Is there any way I can move this along? Someone mentioned the Jenkins Bill of Materials. I have really no idea what you want me to do here. Can I use that to see if I break other plugins, without issuing a new release of WorkflowJob?

Yes exactly that runs the tests of many popular plugins with the version defined in the pom and can be used to pick up the affect of a change on many plugins

GLundh:
> I put the incremental up in the bom latest here: [jenkinsci/bom#468](https://github.com/jenkinsci/bom/pull/468)

There seems to be an error, but not related to this change? Or am I reading that PR wrong?

timja:
> > I put the incremental up in the bom latest here: [jenkinsci/bom#468](https://github.com/jenkinsci/bom/pull/468)
> 
> There seems to be an error, but not related to this change? Or am I reading that PR wrong?

unrelated to the PR, but related to this repo, the plugin needs re-incrementalising, speicifically the `githubRepo` property is missing,

https://github.com/jenkinsci/incrementals-tools

see archetype:
https://github.com/jenkinsci/archetypes/blob/master/empty-plugin/src/main/resources/archetype-resources/pom.xml#L19

GLundh:
> > > I put the incremental up in the bom latest here: [jenkinsci/bom#468](https://github.com/jenkinsci/bom/pull/468)
> > 
> > 
> > There seems to be an error, but not related to this change? Or am I reading that PR wrong?
> 
> unrelated to the PR, but related to this repo, the plugin needs re-incrementalising, specifically the `githubRepo` property is missing,
> 
> https://github.com/jenkinsci/incrementals-tools
> 
> see archetype:
> https://github.com/jenkinsci/archetypes/blob/master/empty-plugin/src/main/resources/archetype-resources/pom.xml#L19

Thanks. Is that something you want me to include in this specific PR?

timja:
See https://github.com/jenkinsci/workflow-job-plugin/pull/185

timja:
Tests passed in bom

jglick:
Relates to #137 and (especially) #131.

PR #199, opened 2021-06-30
Improve durability of `SURVIVABLE_NONATOMIC` while retaining the skipping of `fsync(2)`

basil:
Not yet ready for review. Filing just to get an incremental build.

See [JENKINS-66001](https://issues.jenkins.io/browse/JENKINS-66001). Downstream of jenkinsci/jenkins#5599, jenkinsci/workflow-api-plugin#163, and jenkinsci/workflow-support-plugin#120. Part 5 of a 5-part series to make Pipeline's `SURVIVABLE_NONATOMIC` mode behave [as advertised in the documentation](https://www.jenkins.io/doc/book/pipeline/scaling-pipeline/#what-am-i-giving-up-with-this-durability-setting-trade-off).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,19 +64,20 @@
     <properties>
         <revision>2.42</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.222.4</jenkins.version>
+        <jenkins.version>2.300-rc31329.23a941f5b44f</jenkins.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/5599 -->
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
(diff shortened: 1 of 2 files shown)


jglick:
> I tested this by…

If I get this right, you actually had a real Jenkins installation which had poor performance in the conservative mode but file corruption in the performance mode and could not find a happy medium? So have you tried running incremental builds of the core + plugin patches to verify that it all works the way you really want?

basil:
I don't even run weeklies, let alone incrementals, in my production environment at my company. And my staging environment at my company doesn't get enough traffic to be a realistic load test. But in a synthetic test with a Pipeline job with 100 steps, this took the number of `fsync(2)` calls from 426 to 7. (The remaining ones were things like updating the flow execution list, queue maintenance, etc.) At a high level it makes sense to `fsync(2)` for relatively infrequent "important" things but not for every Groovy continuation of every build.

jglick (approved):

PR #200, opened 2021-06-30
[JENKINS-43353] Option to disable concurrent builds by aborting previous ones

jglick:
[JENKINS-43353](https://issues.jenkins.io/browse/JENKINS-43353), more straightforward and efficient than using the previous hack

```groovy
def buildNumber = BUILD_NUMBER as int; if (buildNumber > 1) milestone(buildNumber - 1); milestone(buildNumber)
```

Can be configured via GUI (for standalone Pipelines) or as code

```groovy
properties([disableConcurrentBuilds(abortPrevious: true)])
```

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
@@ -25,7 +25,14 @@
  */
 package org.jenkinsci.plugins.workflow.job.properties;
 
+import hudson.model.Result;
+import hudson.model.queue.QueueTaskFuture;
+import jenkins.model.InterruptedBuildAction;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
+import static org.junit.Assert.assertEquals;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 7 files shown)


jglick:
Some other usages that could be cleaned up:
* https://github.com/jenkinsci/jenkins/blob/677fbb2a81fa8e9a0bc37ad5eca56565612e1f2e/Jenkinsfile#L8
* https://github.com/jenkinsci/bom/blob/c0d113b152a688f729b52c8551c131b2585348bb/Jenkinsfile#L1
* https://github.com/jenkinsci/plugin-compat-tester/blob/230b64415f1f2500dfd233e983911614f1dd2f3e/Jenkinsfile#L5
* https://github.com/jenkinsci/kubernetes-plugin/blob/d844788b8e186e587ac9cad200ff2dbf3feacc06/Jenkinsfile#L1

src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
```diff
@@ -79,5 +86,62 @@ public void concurrentBuildsMigrationOnByDefault() throws Exception {
...
+        r.waitForCompletion(b2);
+
+        // Control case: simple disable concurrent.
+        p.setConcurrentBuild(false);
+        assertFalse(p.isConcurrentBuild());
+        WorkflowRun b3 = p.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("run/3", b3);
+        SemaphoreStep.success("run/4", null);
+        QueueTaskFuture<WorkflowRun> b4f = p.scheduleBuild2(0);
+        Thread.sleep(1000); // TODO is there a cleaner way for the queue to finish processing?
```
basil, line 124: Perhaps `Queue#maintain` as in https://github.com/jenkinsci/throttle-concurrent-builds-plugin/blob/05f713e9a3c41674b0a963c77e8b2a51e5be6206/src/test/java/hudson/plugins/throttleconcurrents/ThrottleJobPropertyPipelineTest.java#L98 ?

jglick: Afraid not.

```diff
diff --git src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
index 1c19a29..780586f 100644
--- src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
+++ src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
@@ -115,13 +115,11 @@ public class DisableConcurrentBuildsJobPropertyTest {
         r.waitForCompletion(b2);
 
         // Control case: simple disable concurrent.
-        p.setConcurrentBuild(false);
-        assertFalse(p.isConcurrentBuild());
         WorkflowRun b3 = p.scheduleBuild2(0).waitForStart();
         SemaphoreStep.waitForStart("run/3", b3);
         SemaphoreStep.success("run/4", null);
         QueueTaskFuture<WorkflowRun> b4f = p.scheduleBuild2(0);
-        Thread.sleep(1000); // TODO is there a cleaner way for the queue to finish processing?
+        r.jenkins.getQueue().maintain();
         assertFalse(b4f.getStartCondition().isDone());
         SemaphoreStep.success("run/3", null);
         r.waitForCompletion(b4f.waitForStart());
```

falls through to an NPE at https://github.com/jenkinsci/workflow-job-plugin/blob/1a07b9ff3ff4336e020a5076dd72e776ba4845e8/src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java#L131 while

```diff
diff --git src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
index 1c19a29..7c06f85 100644
--- src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
+++ src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
@@ -115,12 +115,11 @@ public class DisableConcurrentBuildsJobPropertyTest {
         r.waitForCompletion(b2);
 
         // Control case: simple disable concurrent.
-        p.setConcurrentBuild(false);
-        assertFalse(p.isConcurrentBuild());
         WorkflowRun b3 = p.scheduleBuild2(0).waitForStart();
         SemaphoreStep.waitForStart("run/3", b3);
         SemaphoreStep.success("run/4", null);
         QueueTaskFuture<WorkflowRun> b4f = p.scheduleBuild2(0);
+        r.jenkins.getQueue().maintain();
         Thread.sleep(1000); // TODO is there a cleaner way for the queue to finish processing?
         assertFalse(b4f.getStartCondition().isDone());
         SemaphoreStep.success("run/3", null);
```

fails at https://github.com/jenkinsci/workflow-job-plugin/blob/1a07b9ff3ff4336e020a5076dd72e776ba4845e8/src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java#L125

This test really needs to check that not only does `Queue` finish its work, but that there is not a new `WorkflowRun` started, which I think is a separate asynchronous process.

jglick: Oh I see, checking `CauseOfBlockage` helps. Working on it…

jglick: Indeed

```diff
diff --git src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
index 1c19a29..88bfa62 100644
--- src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
+++ src/test/java/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest.java
@@ -27,6 +27,10 @@ package org.jenkinsci.plugins.workflow.job.properties;
 
 import hudson.model.Result;
 import hudson.model.queue.QueueTaskFuture;
+import java.util.Collections;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import jenkins.model.BlockedBecauseOfBuildInProgress;
 import jenkins.model.InterruptedBuildAction;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
@@ -115,14 +119,12 @@ public class DisableConcurrentBuildsJobPropertyTest {
         r.waitForCompletion(b2);
 
         // Control case: simple disable concurrent.
-        p.setConcurrentBuild(false);
-        assertFalse(p.isConcurrentBuild());
         WorkflowRun b3 = p.scheduleBuild2(0).waitForStart();
         SemaphoreStep.waitForStart("run/3", b3);
         SemaphoreStep.success("run/4", null);
         QueueTaskFuture<WorkflowRun> b4f = p.scheduleBuild2(0);
-        Thread.sleep(1000); // TODO is there a cleaner way for the queue to finish processing?
-        assertFalse(b4f.getStartCondition().isDone());
+        r.jenkins.getQueue().maintain();
+        assertEquals(Collections.singletonList(BlockedBecauseOfBuildInProgress.class), r.jenkins.getQueue().getItems(p).stream().map(i -> i.getCauseOfBlockage().getClass()).collect(Collectors.toList()));
         SemaphoreStep.success("run/3", null);
         r.waitForCompletion(b4f.waitForStart());
         r.waitForCompletion(b3);
```

fails as expected:

```
expected:<[class jenkins.model.BlockedBecauseOfBuildInProgress]> but was:<[class hudson.model.queue.CauseOfBlockage$NeedsMoreExecutorImpl]>
```

(Well, sort of. I would not have expected a flyweight task to have `NeedsMoreExecutorImpl`; would have expected an empty list or something. But close enough.)

basil (approved):

jglick:
@cathychan & @jtnord point out that when introducing this option for the first time, say to replace the `milestone` trick, you leave one extra build of an existing job running because the property is only interpreted when starting a build, not when the property is added (for example via `properties` step in `Jenkinsfile`). Probably `DisableConcurrentBuildsJobProperty` could be improved to override `setOwner` to immediately abort any running build(s) other than the latest.

PR #201, opened 2021-07-15
Add missing symbol for WorkflowJob

fcojfernandez:
WorkflowJob lacks a `Symbol` annotation. This PR adds the symbold

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -87,6 +87,7 @@ import jenkins.model.ParameterizedJobMixIn;
 import jenkins.model.lazy.LazyBuildMixIn;
 import jenkins.triggers.SCMTriggerItem;
 import net.sf.json.JSONObject;
+import org.jenkinsci.Symbol;
 import org.jenkinsci.plugins.workflow.flow.BlockableResume;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -656,7 +657,9 @@ public static void alias() {
         WorkflowRun.alias();
     }
 
-    @Extension(ordinal=1) public static final class DescriptorImpl extends TopLevelItemDescriptor {
+    @Extension(ordinal=1)
+    @Symbol("pipeline")
```
jtnord, line 661: FreeStyle has both `freeStyle` and `freeStyleJob`   not clear to me what the convention here is, should we be including `"Job"` or not?

jtnord (approved):

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
fcojfernandez: Neither do I know, so I've opted for pipeline since it's popularly known that way. I can add both if desired

rsandell (approved):

PR #235, opened 2022-01-21
JENKINS-67522 Use same animation as core for pipeline logs

timja:
see https://issues.jenkins.io/browse/JENKINS-67552

Introduced in 2.320 in https://github.com/jenkinsci/jenkins/pull/5871

We could do version detection but up to maintainers, this would be the simplest approach

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
@@ -270,7 +270,7 @@ public class WorkflowRunRestartTest {
 
     private boolean hasTermOrKillLink(WorkflowRun b, String termOrKill) throws Exception {
         return !story.j.createWebClient().getPage(b)
-                .getByXPath("//a[@href = '#' and contains(@onclick, '/" + b.getUrl() + termOrKill + "')]").isEmpty();
+                .getByXPath("//a[@href = '#' and contains(@data-url, '/" + b.getUrl() + termOrKill + "')]").isEmpty();
     }
 
     public static class Zombie extends Step {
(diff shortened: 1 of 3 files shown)


jglick (approved):
OK? There are also two spots in `workflow-support-plugin` that would need the same treatment.

car-roll (approved):

timja:
> OK? There are also two spots in `workflow-support-plugin` that would need the same treatment.

There’s quite a few references in plugins but this is the most visible one, others can prob wait a bit longer I think

timja:
FML:
https://github.com/jenkinsci/bom/blob/231c8d6403cf235a1f149fd24db857395d510cad/excludes.txt#L16-L17

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
```diff
@@ -270,7 +270,7 @@
 
     private boolean hasTermOrKillLink(WorkflowRun b, String termOrKill) throws Exception {
         return !story.j.createWebClient().getPage(b)
-                .getByXPath("//a[@href = '#' and contains(@onclick, '/" + b.getUrl() + termOrKill + "')]").isEmpty();
+                .getByXPath("//a[@href = '#' and contains(@data-url, '/" + b.getUrl() + termOrKill + "')]").isEmpty();
```
timja, line 273: Behaviour changed in https://github.com/jenkinsci/jenkins/pull/5515 in 2.320

Not picked up in bom due to https://github.com/jenkinsci/bom/blob/231c8d6403cf235a1f149fd24db857395d510cad/excludes.txt#L16-L17

MarkEWaite:
@timja the title of the pull request includes the wrong issue ID.  You reference the correct issue ID as https://issues.jenkins.io/browse/JENKINS-67552

Could the changelog be corrected to link to the correct issue ID https://issues.jenkins.io/browse/JENKINS-67552?

JENKINS-67522 is the incorrect issue ID

timja:
Sorted thanks

PR #256, opened 2022-04-21
Update plugin parent POM and BOM

basil:


basil:
The changes to local data are because the local data contained unloadable builds which, though unessential to the test, caused the Windows test runs to fail when trying to clean up the Jenkins home directory because of the unloaded builds holding files open. This PR removes the unloadable builds while keeping the job definitions, which is all that the tests needed anyway.

@jenkinsci/workflow-job-plugin-developers A release of this would facilitate Java 17 PCT testing.

jglick (approved):
Tests from #8 & #9. I had to open the ZIPs to see the diff. For `@LocalData` it is typically recommended to unpack the ZIP (thus, e.g., `src/test/resources/org/jenkinsci/plugins/workflow/job/properties/DisableConcurrentBuildsJobPropertyTest/concurrentBuildsMigrationFromFalse/jobs/p/config.xml`) for easier tracking.

basil:
@car-roll Would you be able to help merge and release this?

basil:
Thank you!

PR #260, opened 2022-05-11
Call `FlowExecutionListener.fireResumed` a bit later

jglick:
Print the message about resuming the build, and schedule `AfterRestartTask`, first. Complements https://github.com/jenkinsci/workflow-api-plugin/pull/221.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -565,11 +565,11 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                         }
 
                         if (Boolean.FALSE.equals(completed)) {
-                            // we've been restarted while we were running. let's get the execution going again.
-                            FlowExecutionListener.fireResumed(fetchedExecution);
-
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -565,11 +565,11 @@ private String key() {
...
 
                         if (Boolean.FALSE.equals(completed)) {
-                            // we've been restarted while we were running. let's get the execution going again.
-                            FlowExecutionListener.fireResumed(fetchedExecution);
-
                             getListener().getLogger().println("Resuming build at " + new Date() + " after Jenkins restart");
                             Timer.get().submit(() -> Queue.getInstance().schedule(new AfterRestartTask(WorkflowRun.this), 0)); // JENKINS-31614
+
+                            // we've been restarted while we were running. let's get the execution going again.
+                            FlowExecutionListener.fireResumed(fetchedExecution);
```
dwnusbaum, line 572: Is this just so that the message gets printed first, or are there more significant effects?

jglick: Mainly just to print the message first.

dwnusbaum (approved):

PR #261, opened 2022-05-15
Update project configuration screen

janfaracik:
As part of https://github.com/jenkinsci/jenkins/pull/6485 the project configuration screen has changed, as well as the key to enable/disable a project, this PR updates this plugin to support those new changes.

image

---


timja (commented):
this change will need a `jenkins.version` bump in the current state

janfaracik:
> > Fwiw, we could use the newly introduced pipeline icon on the job selection page too, if desired.
> 
> The circles the icons are placed in are pure visual and don't have any logic to resize larger symbols.
> 
> Bildschirmfoto 2022-05-23 um 17 13 05
> 
> If it's too much work tinkering with the view boxes of the SVG, just stick to the current one for now, I'd say.

Oh boy did not expect that to happen, reverted it :P

basil:
This PR is in draft state because it contains changes that cannot be merged without breaking backward compatibility. By rebasing this PR on top of #262, which does not break backward compatibility, this PR could be moved out of draft state and closer to the finish line. Since it is only when a PR crosses the finish line that the PR benefits end users, I suggest this action be taken.

timja:
> I still think it would be beneficial to rebase this draft PR on top of #262.

done

basil (approved):

janfaracik:
Hoping I've updated this correctly - does this look okay to you @timja?

timja:
> Hoping I've updated this correctly - does this look okay to you @timja?

Somewhat, you should only use the weekly bom if you are tracking every weekly release as a plugin could be updated to a newer weekly version than you require currently.

This also depends on https://github.com/jenkinsci/workflow-job-plugin/pull/279 now so that CI passes

pom.xml
```diff
@@ -73,8 +73,8 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.346.x</artifactId>
-                <version>1382.v7d694476f340</version>
+                <artifactId>bom-2.346.1</artifactId>
```
timja, line 76:
```suggestion
                <artifactId>bom-2.346.x</artifactId>
```

pom.xml
```diff
@@ -74,7 +74,7 @@
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
                 <artifactId>bom-2.346.x</artifactId>
```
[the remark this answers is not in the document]
timja: no, see https://github.com/jenkinsci/workflow-job-plugin/pull/261#issuecomment-1184300068

jglick (approved):
Seems to work. AFAICT this just changes the icons in the **Configure** screen.

basil:
In `workflow-cps-plugin` with …

```xml
diff --git a/pom.xml b/pom.xml
index e984fb5c..4978eb2d 100644
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
-        <jenkins.version>2.346.1</jenkins.version>
+        <jenkins.version>2.360</jenkins.version>
         <no-test-jar>false</no-test-jar>
         <groovy-cps.version>1.32</groovy-cps.version>
         <node.version>12.19.0</node.version>
@@ -74,7 +74,7 @@
         <dependencies>
             <dependency>
-                <artifactId>bom-2.346.x</artifactId>
+                <artifactId>bom-weekly</artifactId>
                 <version>1478.v81d3dc4f9a_43</version>
                 <scope>import</scope>
                 <type>pom</type>
@@ -148,6 +148,7 @@
         <dependency>
             <artifactId>workflow-job</artifactId>
+            <version>1207.ve6191ff089f8</version>
             <scope>test</scope>
         </dependency>
         <dependency>
```

… `mvn clean verify -Dtest=org.jenkinsci.plugins.workflow.cps.CpsFlowDefinitionTest#cpsScriptAdminModification,org.jenkinsci.plugins.workflow.cps.CpsFlowDefinitionTest#cpsScriptAdminConfiguration` passes.

Updating `workflow-job` from 1207.ve6191ff089f8 to 1226.v44f718dcfe1f …

```xml
diff --git a/pom.xml b/pom.xml
index e984fb5c..bc8111c7 100644
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
-        <jenkins.version>2.346.1</jenkins.version>
+        <jenkins.version>2.360</jenkins.version>
         <no-test-jar>false</no-test-jar>
         <groovy-cps.version>1.32</groovy-cps.version>
         <node.version>12.19.0</node.version>
@@ -74,7 +74,7 @@
         <dependencies>
             <dependency>
-                <artifactId>bom-2.346.x</artifactId>
+                <artifactId>bom-weekly</artifactId>
                 <version>1478.v81d3dc4f9a_43</version>
                 <scope>import</scope>
                 <type>pom</type>
@@ -148,6 +148,7 @@
         <dependency>
             <artifactId>workflow-job</artifactId>
+            <version>1226.v44f718dcfe1f</version>
             <scope>test</scope>
         </dependency>
         <dependency>
```

… `mvn clean verify -Dtest=org.jenkinsci.plugins.workflow.cps.CpsFlowDefinitionTest#cpsScriptAdminModification,org.jenkinsci.plugins.workflow.cps.CpsFlowDefinitionTest#cpsScriptAdminConfiguration` fails with the errors seen in https://ci.jenkins.io/job/Tools/job/bom/job/PR-1059/52/

Since the only change in `workflow-job` between 1207.ve6191ff089f8 to 1226.v44f718dcfe1f is #261, I conclude this PR broke these two tests:

- `org.jenkinsci.plugins.workflow.cps.CpsFlowDefinitionTest#cpsScriptAdminModification`
- `org.jenkinsci.plugins.workflow.cps.CpsFlowDefinitionTest#cpsScriptAdminConfiguration`

This was not detected in https://github.com/jenkinsci/bom/pull/1270 but was detected in my experiments with JENKINS-45047.

CC @janfaracik @jglick @timja

jglick:
These tests were added by @yaroslavafenkin in https://github.com/jenkinsci/workflow-cps-plugin/pull/544 and I am not familiar with them. Will try to narrow down the specific cause of regression.

jglick:
It is specifically the first hunk in `WorkflowJob`—the switch to `req.bindJSON`—which causes the regression.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -180,25 +180,15 @@ public void setDefinition(FlowDefinition definition) {
     @Override protected void submit(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException, Descriptor.FormException {
         super.submit(req, rsp);
         JSONObject json = req.getSubmittedForm();
-        // TODO: Switch from Descriptor.newInstancesFromHeteroList to Descriptor.bindJSON when minimum supported
-        //  version is Jenkins 2.342 or higher.
-        definition = json.get("definition") == null ? null :
-                Descriptor.newInstancesFromHeteroList(req, json.getJSONObject("definition"),
-                        ExtensionList.lookup(FlowDefinitionDescriptor.class)).get(0);
+        definition = req.bindJSON(FlowDefinition.class, json.getJSONObject("definition"));
```
yaroslavafenkin, line 183: `StaplerRequest#bindJSON` isn't the same as `Descriptor#bindJSON`

dwnusbaum: Yeah, this either needs to be reverted to match https://github.com/jenkinsci/workflow-job-plugin/commit/a25d15548158461749870d328746e2be482f9e4d or changed to use `Descriptor.bindJSON`. It looks like <a href="https://github.com/jenkinsci/workflow-job-plugin/pull/261/commits/a8af864562879f1dca792a51417a1439c7fb82c7">this merge commit</a> is where the problem was introduced.

jglick: It is not the same but the idea was that https://github.com/jenkinsci/jenkins/pull/5417 could let us simplify code.

From local testing, using `req.bindJSON` works so long as you _also_ make the call to `Descriptor.newInstancesFromHeteroList` (even later). Something is using weird side effects.

dwnusbaum: > Something is using weird side effects.

Yes, the call to `ScriptApproval.configuring` in `CpsFlowDefinition$DescriptorImpl.newInstance` is a side effect.

Either the code can be reverted to use `Descriptor.newInstancesFromHeteroList` or you can switch to `Descriptor.bindJSON` so that `CpsFlowDefinition$DescriptorImpl.newInstance` gets called.

PR #279, opened 2022-07-14
Remove java 8 from CI

timja:
This plugins needs to swap to the weekly line temporarily in https://github.com/jenkinsci/workflow-job-plugin/pull/261

Java 8 has been dropped and the build is failing there (and Jan can't update the Jenkinsfile)


timja:
cc @jglick

car-roll (approved):

jglick:
BTW this should not have been labeled `developer` (which triggers a release); just `chore` (or `tests`) would suffice.

PR #284, opened 2022-07-25
`Descriptor#bindJSON` not `StaplerRequest#bindJSON`

jglick:
https://github.com/jenkinsci/workflow-job-plugin/pull/261#issuecomment-1194258783; test added in https://github.com/jenkinsci/workflow-cps-plugin/pull/544; new API introduced in https://github.com/jenkinsci/jenkins/pull/5417.

~I do not understand _why_ this fixes the regression, but it seems to, so good enough for now.~

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -180,10 +180,7 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements L
     @Override protected void submit(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException, Descriptor.FormException {
         super.submit(req, rsp);
         JSONObject json = req.getSubmittedForm();
-        // TODO using req.bindJSON(FlowDefinition.class, json.getJSONObject("definition")) breaks CpsFlowDefinitionTest
-        definition = json.get("definition") == null ? null :
-                Descriptor.newInstancesFromHeteroList(req, json.getJSONObject("definition"),
-                        ExtensionList.lookup(FlowDefinitionDescriptor.class)).get(0);
+        definition = Descriptor.bindJSON(req, FlowDefinition.class, json.getJSONObject("definition"));
         authToken = hudson.model.BuildAuthorizationToken.create(req);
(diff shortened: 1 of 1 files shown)


basil (commented):
As a temporary hotfix, OK.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -180,7 +180,10 @@ public void setDefinition(FlowDefinition definition) {
     @Override protected void submit(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException, Descriptor.FormException {
         super.submit(req, rsp);
         JSONObject json = req.getSubmittedForm();
-        definition = req.bindJSON(FlowDefinition.class, json.getJSONObject("definition"));
+        // TODO using req.bindJSON(FlowDefinition.class, json.getJSONObject("definition")) breaks CpsFlowDefinitionTest
+        definition = json.get("definition") == null ? null :
+                Descriptor.newInstancesFromHeteroList(req, json.getJSONObject("definition"),
+                        ExtensionList.lookup(FlowDefinitionDescriptor.class)).get(0);
```
dwnusbaum, line 186: Either re-add the comment from https://github.com/jenkinsci/workflow-job-plugin/commit/a25d15548158461749870d328746e2be482f9e4d about switching to `Descriptor.bindJSON` in the future, or do it now since the baseline is already high enough (untested):
```suggestion
        definition = Descriptor.bindJSON(req, FlowDefinition.class, json.get("definition"));
```

dwnusbaum (dismissed):
> I do not understand why this fixes the regression

See the commit message for https://github.com/jenkinsci/workflow-job-plugin/commit/a25d15548158461749870d328746e2be482f9e4d. `StaplerRequest.bindJSON` does not call overrides of `Descriptor.newInstance` on the bound type. `CpsFlowDefinition` has such an override as of https://github.com/jenkinsci/workflow-cps-plugin/commit/6bd4e8bf837ba1df4394efdd003c70bb1b9ac33f.

dwnusbaum (approved):

yaroslavafenkin (approved):
Not tested manually, but looks OK

PR #285, opened 2022-07-27
Fix: Disable restarting stages in jenkinsfile optionally (JENKINS-54250)

pavlovic-ivan:
  - Added new class member to WorkflowJob that will be utilised by new declarative options to enable/disable "restart from stage" button in Jenkins UI
  - https://issues.jenkins.io/browse/JENKINS-54250
  - https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/560
  - https://github.com/jenkins-infra/jenkins.io/pull/5314


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -116,6 +116,7 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements L
     private transient LazyBuildMixIn<WorkflowJob,WorkflowRun> buildMixIn;
     /** @deprecated replaced by {@link DisableConcurrentBuildsJobProperty} */
     private @CheckForNull Boolean concurrentBuild;
+    private Boolean disableRestartFromStage;
```
dwnusbaum, line 119: "Restart from stage" is a Declarative-specific concept, so we do not want to add this kind of option to this plugin, which is more generic. I would probably have https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/560 define a `class DisableRestartFromStage extends InvisibleAction` and then have `Utils.updateJobProperties` add that action to the job when `disableRestartFromStage(true)` is used (or remove it if someone uses `disableRestartFromStage(false)`) instead of setting this field, and then have `RestartDeclarativePipelineAction.isRestartableFromStage` check for the existence of that action.

dwnusbaum (changes requested):
Thanks for the PR!

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
pavlovic-ivan: @dwnusbaum much appreciated for the feedback. i will be looking into this 🙏 please bare with me, as i am contributing to Jenkins for the first time, so i am getting my head around the entire codebase

dwnusbaum: No worries, let me know if you have other questions. I think that https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/560 along with the changes I described above would be a good approach.

pavlovic-ivan: Sounds good, will do, thanks 🙏

pavlovic-ivan:
@dwnusbaum Closing this draft PR as it is not needed. Changes remain only in pipeline-model-definition-plugin

PR #303, opened 2022-10-17
`WorkflowRun.doTerm` and `doKill` need to redirect back to the build

jglick:
Otherwise clicking the sidepanel entries for `term` / `kill` sends you to a blank page.

PR #307, opened 2022-11-01
reintroduce binary compatabilty lost in #303

jtnord:
as yet untest approach to reintroce binary api compatability lost in #303 without the need to update several other plugins.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -84,6 +84,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.servlet.ServletException;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import jenkins.model.CauseOfInterruption;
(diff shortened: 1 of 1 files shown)


jtnord:
@jglick trying to test this manually - but in the original you say 
> Otherwise clicking the sidepanel entries for term / kill sends you to a blank page.

However I do not see any side panel entries for that - so unsure how you are reaching it.

image

image


there is `doStop` from the red x, but I think you are referring to something else?

jglick:
Yes I am referring to a black **X** icon in the sidepanel. https://github.com/jenkinsci/workflow-job-plugin/blob/9b3f35756381efcc95550084320e342e903c8609/src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly#L55-L74 I seem to recall there being some conditions under which these were not shown.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -872,7 +889,8 @@ public synchronized HttpResponse doStop() {
         if (e != null) {
             return e.doStop();
         } else {
-            return doKill();
+            doKill();
+            return HttpResponses.forwardToPreviousPage();
```
jglick, line 893: Does the double response break anything?

jglick (approved):
_Looks_ like it should work.

jglick:
Another approach I thought of was e.g.

```java
public void doTerm() {
  // as previously
}
@RequirePOST
@WebMethod("term") public HttpResponse httpTerm() {
  doTerm();
  return HttpResponses.forwardToPreviousPage();
}
```

if Stapler in fact likes that.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jtnord: 🤔  possibly a stack trace that a response has been committed.  I need to manually confirm :)

jtnord:
tested with the following pipleine

```
while (true) {
    try {
        sleep 500000
    } catch(e) {}
}
```
image


this caused the links to show up in the sidepanel.  they worked without any stack showing in the output of `mvn hpi:run`

I can investigate the alternative if you wish?

PR #308, opened 2022-11-01
reintroduce binary compatabilty lost in #303

jtnord:
alternative to #307 proposed by @jglick 

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -127,8 +127,10 @@ import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.stapler.HttpResponse;
 import org.kohsuke.stapler.HttpResponses;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -475,15 +476,21 @@ public HttpResponse doTerm() {
             @Override public void onFailure(@NonNull Throwable t) {}
         });
         printLater(StopState.KILL, "Click here to forcibly kill entire build");
+    }
+
+    /** Sends {@link StepContext#onFailure} to all running (leaf) steps. */
+    @RequirePOST
+    @WebMethod(name = { "term" })
```
jglick, line 483:
```suggestion
    @WebMethod(name = "term")
```
conventionally

jglick: probably also want to mark `DoNotUse` to make it clear this is here only for use from Stapler

jglick (approved):
Does it work?

jtnord:
> Does it work?

yes

jglick (approved):

PR #352, opened 2023-05-15
Implementing `Job.getEstimatedDurationCandidates`

jglick:
Downstream of https://github.com/jenkinsci/jenkins/pull/7998.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,8 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.361.4</jenkins.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -63,7 +63,7 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.361.4</jenkins.version>
+        <jenkins.version>2.407</jenkins.version>
```
jglick, line 66: Post-LTS version, but OTOH it looks like we will need one for #364 anyway.

jtnord (commented):
should the bom be updated too here (either a weekly 😱  or the closest LTS version)

jtnord (approved):

car-roll (approved):

PR #356, opened 2023-05-23
Switch `WorkflowRunRestartTest` to `JenkinsSessionRule`

jglick:


src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
```diff
@@ -117,8 +118,8 @@ public class WorkflowRunRestartTest {
     }
 
     @Issue({"JENKINS-45585", "JENKINS-50784"})  // Verifies execution lazy-load
-    @Test public void lazyLoadExecution() {
-        story.thenWithHardShutdown(r -> {
```
jglick, line 121: This method does not exist in `JenkinsSessionRule`. Rarely needed; no apparent reason for it to have been added in #93.

jglick: …but need to revisit 18d78f305a4526af9cdf3a7b68eb9caf97c7cfbc from #27.

jglick: Ah, have it. `thenWithHardShutdown` blocks `FlowExecutionList.saveAll`, and `unregister` saves asynchronously, so after a restart an attempt is made to load this supposedly running build and `WorkflowRun.getExecution` is called, which can cause the test to fail. Switching to `then` fixes this, as does simply sleeping momentarily.

car-roll (approved):

dwnusbaum (approved):

PR #357, opened 2023-05-23
Improved `WorkflowRun.finish` timing

jglick:
I found that `BulkChange` from `CpsFlowExecution.notifyListeners` caused `WorkflowRun.completed` to be `false` in `build.xml` at the moment `RunListener.onFinalized` is called. Was noted in 18d78f305a4526af9cdf3a7b68eb9caf97c7cfbc in #27.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -646,6 +646,11 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                 RunListener.fireCompleted(WorkflowRun.this, myListener);
                 fireCompleted();
                 myListener.finished(getResult());
+                try {
+                    StashManager.maybeClearAll(this, myListener);
+                } catch (IOException | InterruptedException x) {
+                    Functions.printStackTrace(x, myListener.error("Failed to clean up stashes"));
+                }
                 if (myListener instanceof AutoCloseable) {
                     try {
                         ((AutoCloseable) myListener).close();
@@ -666,12 +671,6 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
             onEndBuilding();
         } finally {  // Ensure this is ALWAYS removed from FlowExecutionList
             FlowExecutionList.get().unregister(new Owner(this));
-        }
-
-        try {
-            StashManager.maybeClearAll(this, /* or move up before closing getListener()? */new LogTaskListener(LOGGER, Level.FINE));
-        } catch (IOException | InterruptedException x) {
-            LOGGER.log(Level.WARNING, "failed to clean up stashes from " + this, x);
         }
         completeAsynchronousExecution();
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -453,6 +453,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                 }
                 if (modified) {
                     saveWithoutFailing(true);
+                    completeAsynchronousExecution();
                 }
                 return;
             }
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -663,7 +664,6 @@ private void finish(@NonNull Result r, @CheckForNull Throwable t) {
             onEndBuilding();
         } finally {  // Ensure this is ALWAYS removed from FlowExecutionList
             FlowExecutionList.get().unregister(new Owner(this));
-            listener = null;
```
jglick, line 666: Also need to move up `listener` clearing. Otherwise `listener` wold be non-null (thus still `logUpdated`) at the moment `finish` saved, thus failing `WorkflowRunTest.basics`.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunRestartTest.java
```diff
@@ -118,7 +118,7 @@ public class WorkflowRunRestartTest {
 
     @Issue({"JENKINS-45585", "JENKINS-50784"})  // Verifies execution lazy-load
     @Test public void lazyLoadExecution() {
-        story.thenWithHardShutdown(r -> {
+        story.then(r -> {
```
jglick, line 121: See https://github.com/jenkinsci/workflow-job-plugin/pull/356#discussion_r1203051438.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -663,14 +671,8 @@ private void finish(@NonNull Result r, @CheckForNull Throwable t) {
...
-        }
-
-        try {
-            StashManager.maybeClearAll(this, /* or move up before closing getListener()? */new LogTaskListener(LOGGER, Level.FINE));
```
jglick, line 670: Fixes a longstanding issue, that for example https://github.com/jenkinsci/artifact-manager-s3-plugin/blob/b8c98c6978ea3d17517e936b73ac4e98b735e934/src/main/java/io/jenkins/plugins/artifact_manager_jclouds/JCloudsArtifactManager.java#L355 would not be seen.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -663,14 +671,8 @@ private void finish(@NonNull Result r, @CheckForNull Throwable t) {
...
-        } catch (IOException | InterruptedException x) {
-            LOGGER.log(Level.WARNING, "failed to clean up stashes from " + this, x);
         }
+        completeAsynchronousExecution();
```
dwnusbaum, line 675: It seems unusual to have this one line outside of the `try`/`finally` that surrounds the rest of the method. Should it go into the `finally` block in case something in `try` throws? Or is the `try` block not actually necessary in practice and so it does not really matter?

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -748,7 +751,7 @@ private void fireCompleted(){
                     LOGGER.log(Level.WARNING, "Nulling out FlowExecution due to error in build " + this, x);
                     execution = null; // probably too broken to use
                     executionLoaded = true;
-                    saveWithoutFailing(); // Ensure we do not try to load again
+                    saveWithoutFailing(true); // Ensure we do not try to load again
```
dwnusbaum, line 754: Should we call `completeAsynchronousExecution()` here?
```suggestion
                    saveWithoutFailing(true); // Ensure we do not try to load again
                    completeAsynchronousExecution();
```

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Probably makes sense to put it there as well.

dwnusbaum (approved):

PR #371, opened 2023-07-18
Write NewNodeConsoleNote to per-step logs instead of overall log

dwnusbaum:
This is currently just a prototype I am using to investigate some things. Requires https://github.com/jenkinsci/workflow-api-plugin/pull/296 to fix a test that happens to trigger that bug with this change.

I am working on a (proprietary) Pipeline log viewer that allows users to view logs for an entire stage or branch, by combining the per-step logs for all the steps in the stage/branch. This works fine, but there are a few issues:
* You lose the `[Pipeline] <step>` lines from `NewNodeConsoleNote` that give you context about what steps are running, which can be confusing if the stage/branch has a lot of steps.
    * There is no supported way to include only the relevant step announcements using public APIs today, you have to either include or exclude lines from the overall log.
* It is difficult to correctly identify where to put context markers indicating when log output changes from one branch to another (as in the left gutter in the classic console). Given two `[Pipeline] <step>` lines in a row, there is no exact way to mark their context.
* It is very awkward to try to identify "meaningful" lines from the overall log, e.g. Jenkins restarting, exceptions at the end of a build, etc.

Emitting the `NewNodeConsoleNote` lines to the per-step logs instead allows all of these issues to be addressed using public APIs. I think it could be useful to the classic console view and other Pipeline log visualizations as well.

Of course there are some drawbacks:
* Visualizations that only show logs for a single step will show at least 1 new log line, i.e. `[Pipeline] <step name>`, which is at best useless for these visualizations since they already indicate what step is selected in some other way.
* Various changes would need to be made to preserve the show/hide behavior in the classic console view.
* I would not be surprised if this breaks tests in various plugins, but I will check at least all BOM plugins if I decide to pursue this change further.
* Probably some other things I haven't thought of.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -86,6 +86,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.Closeable;
+import java.io.InterruptedIOException;
 import jenkins.model.CauseOfInterruption;
 import jenkins.model.Jenkins;
 import jenkins.model.lazy.BuildReference;
@@ -1080,7 +1082,23 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
      */
     private final class NodePrintListener implements GraphListener.Synchronous {
         @Override public void onNewHead(FlowNode node) {
-            NewNodeConsoleNote.print(node, getListener());
+            // TODO: Breaks show/hide links in the console because the line with the show/hide link gets hidden.
+            LogStorage storage = LogStorage.of(asFlowExecutionOwner());
+            try {
+                TaskListener listener = TaskListenerDecorator.apply(storage.nodeListener(node), asFlowExecutionOwner(), null);
+                try {
+                    NewNodeConsoleNote.print(node, listener);
+                } finally {
+                    if (listener instanceof Closeable) {
+                        ((Closeable) listener).close();
+                    }
+                }
+            } catch (InterruptedIOException | InterruptedException e) {
+                Thread.currentThread().interrupt();
+                LOGGER.log(Level.FINE, e, () -> "Unable to write NewNodeConsoleNote for node " + node.getId());
+            } catch (IOException e) {
+                LOGGER.log(Level.FINE, e, () -> "Unable to write NewNodeConsoleNote for node " + node.getId());
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
```diff
@@ -60,7 +60,7 @@
  * @see LogStorage#startStep
  */
 @Restricted(NoExternalUse.class)
-public class NewNodeConsoleNote extends ConsoleNote<WorkflowRun> {
+public class NewNodeConsoleNote extends ConsoleNote {
```
dwnusbaum, line 63: This issue is kind of interesting. I am not sure if it has been reported previously, but any console notes that rely on a `Run` context totally break the per-step log view in the Pipeline steps view. (The notes aren't just skipped, they cause a `ClassCastException` [here](https://github.com/jenkinsci/jenkins/blob/7fb5f24efd92eda2461e9bf2f1fdc50295e12573/core/src/main/java/hudson/console/ConsoleAnnotationOutputStream.java#L119) and break all log display!)

Perhaps [`LogStorage.stepLog`](https://github.com/jenkinsci/workflow-api-plugin/blob/1679fa2f0f761c285a7309b8781c72e8fd9601c8/src/main/java/org/jenkinsci/plugins/workflow/log/LogStorage.java#L115) should really return a `AnnotatedLargeText<WorkflowRun>` instead? Not sure if that can be changed compatibly at this point. `ConsoleAnnotationOutputStream` at least could probably be made more robust though.

There do not seem to be many implementations that would trigger the issue, see [here](https://github.com/search?type=code&q=+owner%3Ajenkinsci+consolenote%3CWorkflowRun) and [here](https://github.com/search?type=code&q=+owner%3Ajenkinsci+ConsoleNote%3CRun).

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1080,7 +1082,23 @@ private final class GraphL implements GraphListener {
      */
     private final class NodePrintListener implements GraphListener.Synchronous {
         @Override public void onNewHead(FlowNode node) {
-            NewNodeConsoleNote.print(node, getListener());
+            // TODO: Breaks show/hide links in the console because the line with the show/hide link gets hidden.
```
dwnusbaum, line 1085: Should be fixable, but I am not going to bother looking into it unless I really want to get this merged.

src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
jglick: Hmm, does this have anything to do with https://github.com/jenkinsci/jenkins/pull/3662?

dwnusbaum:
For now we are instead going to filter these out when reading the log file for our visualization.

PR #377, opened 2023-08-21
Expose `NewNodeConsoleNote` as `Beta.class` so `LogStorage` implementations can interact with it

dwnusbaum:
`NewNodeConsoleNote` is special in that custom Pipeline log visualizers that are not based on the Jenkins `ConsoleNote`-related APIs likely still want to display the associated lines differently or suppress them entirely (see additional discussion in https://github.com/jenkinsci/workflow-job-plugin/pull/371) rather than just ignoring the markup and showing the raw text content.

For my use case, I can get away with only checking a hard-coded class name against `ConsoleNote.readFrom(...).getClass().getName()` in production code (in fact I can actually reference the class literal due to https://github.com/jenkinsci/lib-access-modifier/issues/166), but I would like to mark this class as `@Restricted(Beta.class)` to make it clearer that API changes should be considered carefully and that incompatible changes should probably be moved to a different class. I would also like to be able to use `NewNodeConsoleNote.print` in tests without having to worry that the API is going to be modified unnecessarily, and to be able to reference the class literal in production code if https://github.com/jenkinsci/lib-access-modifier/issues/166 is fixed.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote.java
@@ -46,6 +46,7 @@ import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.log.LogStorage;
 import org.kohsuke.accmod.Restricted;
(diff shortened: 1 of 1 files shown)


jglick (approved):
Seems reasonable; the bar is low enough for `Beta` and I do not know of any particular plans to rename or delete this class (there is not much signature inside it to change).

PR #391, opened 2023-10-11
Override `isInQueue` & `getQueueItem` from `WorkflowJob`

jglick:
Just noticed that these have placeholder implementations (`false` / `null`) in `Job` which are overridden by `AbstractProject` (because it is a `Queue.Task`) and `WorkflowJob` should do the same.


julieheard (approved):

jglick:
Need review from someone with write access: @olamy @rsandell @dwnusbaum @car-roll

dwnusbaum (approved):
Just curious, what is the expected impact of this change?

jglick:
Good question. I searched for usages of these `Job` methods in core (excluding calls from inside `AbstractProject` itself) including Jelly and AFAICT it only affects the REST API (`@Exported`).

PR #463, opened 2024-09-06
fix(delete): [JENKINS-70252] delete WorkflowRun before WorkflowJob

a1k0u:
We use the repository scanning feature in the cloudbees-folder-plugin with Multi Branch Pipeline project to delete non-existent branches. When the folder plugin finds out which branch should be deleted, it calls the delete method in the `I` class (`AbstractFolder<I>`), in our case it is WorkflowJob. WorkflowJob only overrides the `performDelete` method, where it disables the job, exits the queue and calls `AbstractItem.performDelete()`. This way WorkflowJob only recursively deletes files under it. There is no logic for deleting artifacts under a WorkflowRun. 

But if you delete only the WorkflowRun (_for example `logRotator` in pipeline options_), it will delete artifacts because the parent class calls `deleteArtifacts` in the `delete` method (*fisrt case in the linked issue*). This is why, for example, artifacts are not deleted from S3 after deleting a branch (*second case in the linked issue*).

I believe it would be correct to update `hudson.model.Job`, where to add logic to remove builds (`hudson.model.Run`) from `getBuilds`. But it was much easier to make changes to the plugin and test it in a production environment.

### Testing done

I really don't know how to test this properly. I just made the changes and tested it in our Jenkins instance. Everything is working as I expected. I will be glad to know how to test this fix.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -663,6 +663,11 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements L
     @Override protected void performDelete() throws IOException, InterruptedException {
         setDisabled(true);
         Jenkins.get().getQueue().cancel(this);
+
+        for (WorkflowRun workflowRun : getBuilds()) {
+            workflowRun.delete();
+        }
+
         // TODO call SCM.processWorkspaceBeforeDeletion
         super.performDelete();
     }


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -663,6 +663,11 @@ public void replaceAction(@NonNull Action a) {
     @Override protected void performDelete() throws IOException, InterruptedException {
         setDisabled(true);
         Jenkins.get().getQueue().cancel(this);
+
+        for (WorkflowRun workflowRun : getBuilds()) {
```
jglick, line 667: The trouble is that this will make performance worse for those not using `artifact-manager-s3` (or those using it in its default and only supported mode, which never deletes artifacts), since loading potentially thousands of `WorkflowRun` objects from so many little XML files is expensive. Anyway deleting those S3 directories one by one would be inefficient.

Fixing this properly would have to take a different approach I think. https://javadoc.jenkins.io/jenkins/model/ArtifactManager.html#delete() cannot handle this since the whole object is scoped per build. https://javadoc.jenkins.io/jenkins/model/ArtifactManagerFactory.html could have a `delete(Job)` method, perhaps. Or `JCloudsArtifactManager` (or something else in the plugin) could check `ItemListener.onDeleted` and delete the whole directory with one API call.

jglick (changes requested):

PR #468, opened 2024-09-26
[JENKINS-73824] Add test for deleting a Pipeline job while one of its builds is running only on a OneOffExecutor

dwnusbaum:
See https://github.com/jenkinsci/jenkins/pull/9790. Once I have an incremental build of that PR I will update this PR to consume it.

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.85</version>
+        <version>4.88</version>
         <relativePath/>
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -63,7 +63,11 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.454</jenkins.version>
+        <!-- TODO: Waiting for https://github.com/jenkinsci/workflow-job-plugin/pull/468 to make it to an LTS line. -->
+        <jenkins.version>2.479-rc35394.6ec50c7e34f9</jenkins.version>
+        <!-- Waiting for a release of https://github.com/jenkinsci/plugin-pom/pull/1004 -->
+        <jenkins-test-harness.version>2289.vfd344a_6d1660</jenkins-test-harness.version>
+        <hpi-plugin.version>3.58</hpi-plugin.version>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
         <hpi.compatibleSinceVersion>2.26</hpi.compatibleSinceVersion>
@@ -73,14 +77,21 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.452.x</artifactId>
-                <version>3023.v02a_987a_b_3ff9</version>
+                <artifactId>bom-weekly</artifactId>
+                <version>3387.v0f2773fa_3200</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
         </dependencies>
     </dependencyManagement>
     <dependencies>
+        <dependency>
+            <!-- Waiting for a release of https://github.com/jenkinsci/plugin-pom/pull/1004 -->
+            <groupId>jakarta.servlet</groupId>
+            <artifactId>jakarta.servlet-api</artifactId>
+            <version>5.0.0</version>
+            <scope>provided</scope>
+        </dependency>
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
             <artifactId>ionicons-api</artifactId>
--- a/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
@@ -20,19 +20,24 @@ import hudson.plugins.git.GitSCM;
 import hudson.security.WhoAmI;
 import hudson.triggers.SCMTrigger;
 import java.util.Objects;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import jenkins.plugins.git.GitSampleRepoRule;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition;
+import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
+import org.jvnet.hudson.test.BuildWatcher;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.RunLoadCounter;
 
-
 public class WorkflowJobTest {
+    private static final Logger LOGGER = Logger.getLogger(WorkflowJobTest.class.getName());
+
+    @ClassRule public static BuildWatcher watcher = new BuildWatcher();
 
     @Rule public JenkinsRule j = new JenkinsRule();
     @Rule public GitSampleRepoRule sampleRepo = new GitSampleRepoRule();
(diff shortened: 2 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
```diff
@@ -157,4 +164,38 @@ public void newBuildsShouldNotLoadOld() throws Throwable {
...
+        // Make sure that the job really has been deleted.
+        assertThat(j.jenkins.getItemByFullName(p.getFullName()), nullValue());
+        // ItemDeletion.cancelBuildsInProgress should guarantee that the queue is empty at this point.
+        // (but right now test0 #1 will still be running)
+        var executables = Stream.of(j.jenkins.getComputers())
+                .flatMap(c -> c.getAllExecutors().stream())
+                .map(Executor::getCurrentExecutable)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+        assertThat(executables, empty());
```
dwnusbaum, line 197: Currently fails here, because the build is still running!

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
```diff
@@ -25,6 +31,7 @@
 import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.RunLoadCounter;
 
+
```
jglick, line 34:
```suggestion
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
```diff
@@ -157,4 +164,38 @@ public void newBuildsShouldNotLoadOld() throws Throwable {
         });
     }
 
+    @Test
```
jglick, line 167:
```suggestion
    @Issue("JENKINS-73824")
    @Test
```

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
```diff
@@ -157,4 +164,38 @@ public void newBuildsShouldNotLoadOld() throws Throwable {
...
+    public void deletionShouldWaitForBuildsToComplete() throws Throwable {
+        var p = j.createProject(WorkflowJob.class);
+        p.setDefinition(new CpsFlowDefinition(
+                    "import java.time.Duration\n" +
+                    "import java.time.Instant\n" +
+                    "@NonCPS def tightLoop() {\n" +
```
jglick, line 173: Is this really necessary? Did you try something simpler (and sandbox-friendly) like

```groovy
try {
  sleep 999
} catch (x) {
  echo "aborting soon after $x"
  sleep 10
}
```

?

jglick (approved):

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowJobTest.java
dwnusbaum: Not 100% sure. I think that would have worked as expected prior to the bug being fixed, but now that it is fixed, https://github.com/jenkinsci/jenkins/blob/435bb7989126d6ef3e3d78ddc5597354b71bff74/core/src/main/java/jenkins/model/queue/ItemDeletion.java#L274 will be called repeatedly and so the second sleep will be aborted. Maybe that's ok though, since it means the test will complete more quickly. I'll test it.

jglick: Hmm, not sure why that is called repeatedly.

Note that a single call to https://javadoc.jenkins.io/hudson/model/Executor.html#doStop() (for example, clicking red **X**) currently only sends a single interruption to the program, and `CpsFlowExecution` does not redeliver this in case cleanup drags on. Perhaps `workflow-cps` should automatically retry until the build actually aborts? Or should `workflow-job` automatically escalate to `doTerm` and `doKill`?

dwnusbaum: > not sure why that is called repeatedly.

Yes, it is not clear to me either. It was introduced in https://github.com/jenkinsci/jenkins/pull/2789/commits/047e849d09a914ec0178d98e6f6be681e0b213e5, so I will try to take a closer look at that test, but it seems unrelated to Pipeline.

> Perhaps workflow-cps should automatically retry until the build actually aborts? Or should workflow-job automatically escalate to doTerm and doKill?

I also thought about this and wasn't sure. My first instinct is to just fix the immediate issue, and then if we start seeing problems where jobs can't be deleted because builds are not completing quickly enough, we can consider changes like that.

dwnusbaum: One other non-optimal thing about repeated interruptions with the fix is that you can get a _lot_ of [these tasks](https://github.com/jenkinsci/workflow-job-plugin/blob/d82e9f82c8115f04fa3f27fc38e18ca01c6410d2/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L385-L403) if interruption is not immediate, and I could see it causing system log/build log spam. I didn't have a `BuildWatcher` active previously though, so let me check it again to see if that message gets printed a bunch.

dwnusbaum: > let me check it again to see if that message gets printed a bunch.

The 15 second timer in `printLater` makes that not a problem. From some logging though I can see that around 200 timer tasks get submitted by calls to `interrupt` for a build that takes 10 seconds to stop. I will look at why that `interruptOnDelete` test required repeated calls to `interrupt` because it seems like we don't really want to repeatedly call interrupt like this.

dwnusbaum: Hmm, well `JobTest.interruptOnDelete` passes even if the extra interruptions are removed. Now I am leaning towards removing them, even just for the sake of consistency with other interruption mechanisms in Jenkins.

dwnusbaum: As far as the original question though, yes we can simplify the Pipeline here, but we need two back-to-back calls to `sleep` in the catch block for the issue to reproduce consistently without the fix.

dwnusbaum: And use https://github.com/jenkinsci/jenkins/pull/9790#discussion_r1777792002 for further discussion about repeated interruption.

dwnusbaum: > As far as the original question though, yes we can simplify the Pipeline here, but we need two back-to-back calls to sleep in the catch block for the issue to reproduce consistently without the fix.

In practice actually it seems unnecessary. [3426a6a](https://github.com/jenkinsci/workflow-job-plugin/pull/468/commits/3426a6ac25629e3dba064468f9e3cac9280b0a0c) simplifies the test.

dwnusbaum:
The CodeQL autobuild doesn't like the HPI plugin override (interestingly I also get this error in Netbeans, but it works fine using Maven directly):

>  [2024-09-27 19:27:07] [autobuild] org.apache.maven.project.ProjectBuildingException: Some problems were encountered while processing the POMs:
  Error: 9-27 19:27:07] [autobuild] [ERROR] Unknown packaging: hpi @ org.jenkins-ci.plugins.workflow:workflow-job:${changelist}, /home/runner/work/workflow-job-plugin/workflow-job-plugin/pom.xml, line 37, column 16

I suspect that a release of https://github.com/jenkinsci/plugin-pom/pull/1004 would help, but I have not tried to investigate the issue deeply.

basil:
https://github.com/jenkins-infra/jenkins-security-scan/pull/32 FTR

pom.xml
```diff
@@ -73,14 +74,21 @@
...
         </dependencies>
     </dependencyManagement>
     <dependencies>
+        <dependency>
+            <!-- Waiting for a release of https://github.com/jenkinsci/plugin-pom/pull/1004 -->
+            <groupId>jakarta.servlet</groupId>
+            <artifactId>jakarta.servlet-api</artifactId>
+            <version>5.0.0</version>
+            <scope>provided</scope>
+        </dependency>
```
basil, line 91: No longer needed as of #473.

```suggestion
```

pom.xml
```diff
@@ -63,7 +63,8 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.479</jenkins.version>
+        <!-- TODO: Waiting for https://github.com/jenkinsci/workflow-job-plugin/pull/468 to make it to an LTS line. -->
```
basil, line 66:
```suggestion
        <!-- TODO: Waiting for https://github.com/jenkinsci/jenkins/pull/9790 to make it to an LTS line. -->
```

pom.xml
```diff
@@ -73,8 +74,8 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.462.x</artifactId>
-                <version>3221.ve8f7b_fdd149d</version>
+                <artifactId>bom-weekly</artifactId>
```
basil, line 77: Per https://github.com/jenkinsci/bom?tab=readme-ov-file#depending-on-bom-weekly

```suggestion
                <artifactId>bom-2.462.x</artifactId>
```

jglick:
Is there any reason for this to still be in draft?

pom.xml
```diff
@@ -63,7 +63,7 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.479</jenkins.version>
+        <jenkins.version>2.480</jenkins.version>
```
dwnusbaum, line 66: Given this just adds a test, I am not sure if we actually want to merge this as-is, or would prefer to wait for JENKINS-73824 to possibly get backported to 24.79.x.

jglick: Oh OK, it is an `lts-candidate`.

dwnusbaum:
Integrated into https://github.com/jenkinsci/workflow-job-plugin/pull/470 for simplicity.

PR #470, opened 2024-10-01
[JENKINS-73824][JENKINS-73835] Remove redundant log rotation after changes in core and add regression tests related to deleting Pipeline jobs and builds

dwnusbaum:
See [JENKINS-73835](https://issues.jenkins.io/browse/JENKINS-73835) and https://github.com/jenkinsci/jenkins/pull/9810. Thematically related to https://github.com/jenkinsci/workflow-job-plugin/pull/468 and https://github.com/jenkinsci/jenkins/pull/9790.

The upstream changes allow us to remove the direct call to `getParent().logRotate()` from `WorkflowRun.finish()` in this PR, and we can add a regression test for log rotation potentially cleaning up Pipeline builds too early.

### Testing done

See new automated test.

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
         <!-- TODO: Waiting for https://github.com/jenkinsci/jenkins/pull/9810 to make it to an LTS line. -->
-        <jenkins.version>2.479-rc35393.97fb_6a_a_25d9b_</jenkins.version>
+        <jenkins.version>2.479-rc35395.f676b_e52239c</jenkins.version>
         <!-- Waiting for a release of https://github.com/jenkinsci/plugin-pom/pull/1004 -->
         <jenkins-test-harness.version>2289.vfd344a_6d1660</jenkins-test-harness.version>
         <hpi-plugin.version>3.58</hpi-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -74,8 +74,8 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.462.x</artifactId>
-                <version>3221.ve8f7b_fdd149d</version>
+                <artifactId>bom-2.479.x</artifactId>
+                <version>3482.vc10d4f6da_28a_</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.85</version>
+        <version>4.88</version>
         <relativePath/>
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -63,7 +63,11 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.454</jenkins.version>
+        <!-- TODO: Waiting for https://github.com/jenkinsci/jenkins/pull/9810 to make it to an LTS line. -->
+        <jenkins.version>2.479-rc35393.97fb_6a_a_25d9b_</jenkins.version>
+        <!-- Waiting for a release of https://github.com/jenkinsci/plugin-pom/pull/1004 -->
+        <jenkins-test-harness.version>2289.vfd344a_6d1660</jenkins-test-harness.version>
+        <hpi-plugin.version>3.58</hpi-plugin.version>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
         <hpi.compatibleSinceVersion>2.26</hpi.compatibleSinceVersion>
@@ -73,14 +77,21 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.452.x</artifactId>
-                <version>3023.v02a_987a_b_3ff9</version>
+                <artifactId>bom-weekly</artifactId>
+                <version>3387.v0f2773fa_3200</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
         </dependencies>
     </dependencyManagement>
     <dependencies>
+        <dependency>
+            <!-- Waiting for a release of https://github.com/jenkinsci/plugin-pom/pull/1004 -->
+            <groupId>jakarta.servlet</groupId>
+            <artifactId>jakarta.servlet-api</artifactId>
+            <version>5.0.0</version>
+            <scope>provided</scope>
+        </dependency>
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
             <artifactId>ionicons-api</artifactId>
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -660,13 +660,6 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                 listener = null;
             }
             saveWithoutFailing(true);
-            Timer.get().submit(() -> {
-                try {
-                    getParent().logRotate();
-                } catch (Exception x) {
-                    LOGGER.log(Level.WARNING, "failed to perform log rotation after " + this, x);
-                }
-            });
             onEndBuilding();
         } finally {  // Ensure this is ALWAYS removed from FlowExecutionList
             FlowExecutionList.get().unregister(new Owner(this));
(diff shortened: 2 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -660,13 +660,6 @@ private void finish(@NonNull Result r, @CheckForNull Throwable t) {
                 listener = null;
             }
             saveWithoutFailing(true);
-            Timer.get().submit(() -> {
-                try {
-                    getParent().logRotate();
-                } catch (Exception x) {
-                    LOGGER.log(Level.WARNING, "failed to perform log rotation after " + this, x);
-                }
-            });
```
dwnusbaum, line 669: Originally made asynchronous in https://github.com/jenkinsci/workflow-job-plugin/commit/63fdbe8c2d0f74f97e8f3b92cc566e2fff52bdf3. Note though that since https://github.com/jenkinsci/jenkins/pull/4368, Jenkins core was _also_ calling `Job.logRotate` synchronously from `Run.onEndBuilding` -> `RunListener.onFinalized` (at least with the default global build discarder configuration), so this call was redundant.

With this PR only Jenkins core calls `Job.logRotate`, and it does so synchronously. We could make that call asynchronous if there are any concerns.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -600,6 +604,20 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
...
+            assertThat("Unexpected baseline for checkout at index " + i, checkouts.get(i).pollingBaseline, indexedMatchers[i]);
+        }
+    }
+
+    private static String checkoutString(GitSampleRepoRule repo, boolean changelog, boolean polling) {
+        return "    checkout(changelog:" + changelog +", poll:" + polling +
+                ", scm: [$class: 'GitSCM', branches: [[name: '*/master']], " +
+                ", userRemoteConfigs: [[url: $/" + repo.fileUrl() + "/$]]])\n";
+    }
+
```
dwnusbaum, line 620: Just moving these next to the relevant test for clarity.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -520,7 +524,7 @@ private void assertCulprits(WorkflowRun b, String... expectedIds) throws IOExcep
         assertThat(await().until(() -> ExtensionList.lookupSingleton(CheckCompletedFlag.class).buildXml.get(b.getExternalizableId()), notNullValue()),
             containsString("<completed>true</completed>"));
     }
-    @TestExtension public static final class CheckCompletedFlag extends RunListener<WorkflowRun> {
+    @TestExtension("completedFlag") public static final class CheckCompletedFlag extends RunListener<WorkflowRun> {
```
dwnusbaum, line 527: Only need by this specific test and causes annoying log spam in the new test since it tries to look at build directories for deleted builds.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -667,19 +685,42 @@ public void onInitialize(Run run) {
...
+        for (int i = 0; i < buildsToRun; i++) {
+            r.waitForCompletion(builds[i]);
+        }
+        LOGGER.info("Checking that all build directories are empty ");
+        for (int i = 0; i < buildsToRun; i++) {
+            String[] filesInBuildDir = buildDirs[i].list();
+            if (filesInBuildDir == null) {
+                filesInBuildDir = new String[0];
+            }
+            assertThat("Expected " + buildDirs[i] + " to be empty but saw: " + Arrays.toString(filesInBuildDir), filesInBuildDir, emptyArray());
```
dwnusbaum, line 722: Prior to the fix, this fails intermittently because some directory contains `build.xml`.

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -667,19 +685,42 @@ public void onInitialize(Run run) {
...
-        for (int i = 0; i < checkouts.size(); i++) {
-            assertThat("Unexpected baseline for checkout at index " + i, checkouts.get(i).pollingBaseline, indexedMatchers[i]);
+    @Issue("JENKINS-73835")
+    @Test public void logRotationOnlyProcessesCompletedBuilds() throws Throwable {
+        logging.record(LogRotator.class, Level.FINER);
+        var p = r.createProject(WorkflowJob.class);
+        p.setDefinition(new CpsFlowDefinition(
+                "echo params.FOO; semaphore 'wait'", true));
+        p.addProperty(new ParametersDefinitionProperty(List.of(new StringParameterDefinition("FOO"))));
+        // Keep 0 builds, i.e. delete all builds immediately.
```
dwnusbaum, line 695: I am keeping 0 builds rather than 1 build to simplify test assertions. Mainly I am trying to avoid complicated issues involving the exact timing of calls to `RunList.size` and `RunList.sublist` in `LogRotator.perform` compared to other builds completing, which can cause builds other than the one with the largest build number to be preserved by `LogRotator`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: (#70)

pom.xml
```diff
@@ -63,7 +63,11 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.454</jenkins.version>
+        <!-- TODO: Waiting for https://github.com/jenkinsci/jenkins/pull/9810 to make it to an LTS line. -->
+        <jenkins.version>2.479-rc35393.97fb_6a_a_25d9b_</jenkins.version>
+        <!-- Waiting for a release of https://github.com/jenkinsci/plugin-pom/pull/1004 -->
```
jglick, line 68: https://github.com/jenkinsci/plugin-pom/pull/1004

src/test/java/org/jenkinsci/plugins/workflow/job/WorkflowRunTest.java
```diff
@@ -667,19 +685,42 @@ public void onInitialize(Run run) {
...
+            SemaphoreStep.waitForStart("wait/" + (i+1), b);
+        }
+        for (int i = 0; i < buildsToRun; i++) {
+            SemaphoreStep.success("wait/" + (i+1), null);
+        }
+        LOGGER.info("Waiting for all builds to complete");
+        for (int i = 0; i < buildsToRun; i++) {
+            r.waitForCompletion(builds[i]);
+        }
+        LOGGER.info("Checking that all build directories are empty ");
```
jglick, line 716:
```suggestion
        LOGGER.info("Checking that all build directories are empty");
```

jglick (approved):

dwnusbaum:
Core PR has been merged but not yet released. I think in this case we should go ahead and release the cleanup here, even if it means we need to use a weekly as a baseline temporarily. I think the core fix is a reasonable backporting candidate as well, so I will mark it as `lts-candidate`, and then we could swap to the LTS line here if it does get backported.

pom.xml
```diff
@@ -63,7 +63,8 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.479</jenkins.version>
+        <!-- TODO: Waiting for JENKINS-73824 and JENKINS-73835 to make it into an LTS line -->
+        <jenkins.version>2.481</jenkins.version>
```
dwnusbaum, line 67: Up to reviewers what we want to do here. The removal of the call to `logRotate` in `WorkflowRun.finish` eliminates a redundancy, but things should be fine for 2.481+ users even with it in place. We can either merge this as-is or wait for the fixes to possibly get backported.

pom.xml
```diff
@@ -77,21 +74,14 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-weekly</artifactId>
-                <version>3387.v0f2773fa_3200</version>
+                <artifactId>bom-2.462.x</artifactId>
```
jglick, line 77: https://github.com/jenkinsci/bom/releases/tag/3482.vc10d4f6da_28a_ can now use the 2.479.x BOM.

jglick (approved):
I see no problem with merging as is (just bump the BOM). Can always drop the core dep back down to 2.479.x later if the requisite core fixes are backported.

pom.xml
dwnusbaum: Done in https://github.com/jenkinsci/workflow-job-plugin/pull/470/commits/1232095a5da8aa49b00ed6902dac7dbafb20c508.

PR #472, opened 2024-10-04
`FlowExecution.getOwner` left null after `WorkflowRun.reload`

jglick:
Would not normally matter I think, but can cause confusing error messages from `CpsFlowExecution` which normally assumes that its `owner` is non-null at almost all times; in particular `toString` then makes no mention of which build is having issues.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -553,6 +553,10 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
 
         // super.reload() forces result to be FAILURE, so working around that
         new XmlFile(XSTREAM,new File(getRootDir(),"build.xml")).unmarshal(this);
+        var _execution = execution;
+        if (_execution != null) {
+            _execution.onLoad(new Owner(this));
+        }
     }
 
     @Override protected void onLoad() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -553,6 +553,10 @@ private String key() {
 
         // super.reload() forces result to be FAILURE, so working around that
         new XmlFile(XSTREAM,new File(getRootDir(),"build.xml")).unmarshal(this);
+        var _execution = execution;
```
dwnusbaum, line 556: For what it's worth, my first thought is to instead null out `execution` and reset `executionLoaded` and let them be reloaded by `getExecution` only if needed.

If you prefer to keep the current approach, I would add a `synchronized (this)` block to match `WorkflowRun.getExecution` (unless perhaps typical callers already acquire the monitor for `WorkflowRun` in this case).

jglick: > null out `execution` and reset `executionLoaded`

:thinking: that may work better. Will check it when I get a moment.

jglick: It does not work; `getExecution` simply returns null in that case.

jglick:
This seems to contradict the `lazyLoadExecution` test from #93 :eyes:

jglick:
`WorkflowRunTest.culprits` still sometimes hanging, not sure why.

dwnusbaum (approved):

dwnusbaum (approved):

PR #477, opened 2024-10-08
[JENKINS-73875] CSP compatibility for `NewNodeConsoleNote`

shlomomdahan:
Created a ticket in JIRA to track this issue: [JENKINS-73875](https://issues.jenkins.io/browse/JENKINS-73875)

Removed inline onClick handler from script.js to comply with Content Security Policy and replaced with event delegation.

### Testing done

STATUS QUO: No changes implemented. This is the intended functionality.

https://www.loom.com/share/da8a1ab0f1d8431e858da2ae3ac7215f?sid=2dedd5ac-eb73-472b-a0ae-11e39daff80c

NON RESTRICTIVE: This is the behavior after implementing the CSP changes. The code is working just as before and there is no regression in functionality.

https://www.loom.com/share/1f9479eeb2cf4829bdcf524d6f4d1a37?sid=cfd60732-705b-4874-9ae0-97ae1747b2d2

RESTRICTIVE: This is with CSP restrictive mode enabled. It is working correctly as intended without any issues.

https://www.loom.com/share/cad50c0d86af4e5d9ea467152a54ab08

### Submitter checklist

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
@@ -20,9 +20,9 @@ Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e)
         const toggles = e.querySelectorAll('.pipeline-show-hide .pipeline-toggle');
         const toggle = toggles[toggles.length - 1];
 
-        toggle.addEventListener('click', function(event) {
+        toggle.addEventListener('click', (event) => {
             event.preventDefault();
-            showHidePipelineSection(this);
+            showHidePipelineSection(toggle);
         });
     }
 
--- a/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
@@ -18,12 +18,11 @@ Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e)
         // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
 
         const toggle = e.querySelector('.pipeline-toggle');
-        if (toggle) {
-            toggle.addEventListener('click', function(event) {
-                event.preventDefault();
-                showHidePipelineSection(this);
-            });
-        }
+
+        toggle.addEventListener('click', function(event) {
+            event.preventDefault();
+            showHidePipelineSection(this);
+        });
     }
 
     // The CSS rule for branch names only needs to be added once per node, so we
--- a/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
@@ -15,14 +15,13 @@ Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e)
     var startId = e.getAttribute('startId')
     if (startId == null || startId == nodeId) {
         e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" class="pipeline-toggle">hide</a>)</span>')
-        // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
         const toggles = e.querySelectorAll('.pipeline-show-hide .pipeline-toggle');
         const toggle = toggles[toggles.length - 1];
-
         toggle.addEventListener('click', (event) => {
             event.preventDefault();
             showHidePipelineSection(toggle);
         });
+        // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
     }
     // The CSS rule for branch names only needs to be added once per node, so we
     // check in case we are viewing the truncated log and have already processed
--- a/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
@@ -17,7 +17,7 @@ Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e)
         e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" class="pipeline-toggle">hide</a>)</span>')
         // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
 
-        const toggle = e.querySelector('.pipeline-toggle');
+        const toggle = e.querySelector('.pipeline-show-hide:last-child .pipeline-toggle');
 
         toggle.addEventListener('click', function(event) {
             event.preventDefault();
--- a/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
@@ -16,7 +16,16 @@ Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e)
     if (startId == null || startId == nodeId) {
         e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" class="pipeline-toggle">hide</a>)</span>')
         // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
+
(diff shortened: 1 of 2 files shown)


basil (commented):
I checked some other plugins in `jenkinsci` and found that the typical pattern is to iterate over all elements of a given class with e.g. `document.querySelectorAll` and then add click handlers to just those elements. That seems a little cleaner than adding a click listener on the body element which is a no-op in the vast majority of cases. For example, from `jenkins/core/src/main/resources/hudson/PluginManager/_installed.js`:

```javascript
  document.addEventListener("DOMContentLoaded", function () {
    document
      .querySelectorAll(".plugin-manager-toggle-switch")
      .forEach(function (toggle) {
        toggle.addEventListener("click", flip);
      });
  });
```

shlomomdahan:
Hi @basil, I am running into an issue when trying to do it as you described above.

The links to for the toggle do not appear directly after DOMContentLoaded is triggered. For example, in the below console logs, you can see that those elements don't actually appear in the html tree until all of the dependent resources are loaded.

DOMContentLoaded: Number of .pipeline-toggle elements: 0
Window load: Number of .pipeline-toggle elements: 6 

What do you think is the best approach in this situation?

is it to use `window.addEventListener("load", function() {}`

or perhaps continue to use the event delegation approach but narrow down the parent element to the one that contains the toggles instead of the entire `body`?

Thank you

basil:
Not sure offhand, but the event handler at the top of the file (the third argument to `Behaviour.specify` seems to run every time a `span.pipeline-new-node` is added to the DOM (and itself iterates over all existing nodes with `document.querySelectorAll('.pipeline-new-node')`) so perhaps that would be a good place to fish out the `<a>` element we are interested in and manipulate its onclick handler. We're adding the `<a>` element by setting `e.innerHTML` on line 17, so perhaps we can just add the click event handler right after that?

shlomomdahan:
@basil 
That makes sense. thank you. 

https://www.loom.com/share/d8f0b84fca3f4b218f7d656f268394eb?sid=8b5e1d8e-5de0-43b3-a511-ba3c3737823e

It is working correctly with the new change

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -14,9 +14,18 @@ Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e)
     var nodeId = e.getAttribute('nodeId')
     var startId = e.getAttribute('startId')
     if (startId == null || startId == nodeId) {
-        e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" onclick="showHidePipelineSection(this); return false">hide</a>)</span>')
+        e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" class="pipeline-toggle">hide</a>)</span>')
         // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
+
+        const toggle = e.querySelector('.pipeline-toggle');
+        if (toggle) {
```
basil, line 21: Why would this element ever not be present if we just added it a few lines earlier? I cannot see the purpose of this `if` statement.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -14,9 +14,18 @@ Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e)
     var nodeId = e.getAttribute('nodeId')
     var startId = e.getAttribute('startId')
     if (startId == null || startId == nodeId) {
-        e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" onclick="showHidePipelineSection(this); return false">hide</a>)</span>')
+        e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" class="pipeline-toggle">hide</a>)</span>')
         // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
+
+        const toggle = e.querySelector('.pipeline-toggle');
```
basil, line 20: Are you sure this is always matching the `<span class="pipeline-show-hide">` that we just added on line 17 above? That code is appending `<span class="pipeline-show-hide">` elements as children of `e`, so I am wondering if there might ever be two such children or if there is always one as in your test. Did you try testing nested `parallel` blocks?

shlomomdahan: @basil 

using the below script configuration 

```
parallel first: {
    stage('First Nested') {
        parallel nestedA: {}, nestedB: {}
    }
}, second: {
    stage('"full" details') {
        parallel nestedC: {}, nestedD: {}
    }
}
```

I recorded the behavior of the toggle links:

https://www.loom.com/share/2acb8c6b041a449793132e3f4a433d70


Everything seems to be working okay. Can you confirm if this is the intended behavior?

basil: OK, I am glad the testing is passing. To make this bulletproof, I think we should select the last child with the `pipeline-show-hide` class rather than the first. Since the `/.+/` regular expression is always appending the new `<span>` element to the end of the list of child nodes, selecting the last child should always match the node we just added.

shlomomdahan: Thanks for explaining that detail. New commit addresses this

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -14,9 +14,17 @@ Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e)
     var nodeId = e.getAttribute('nodeId')
     var startId = e.getAttribute('startId')
     if (startId == null || startId == nodeId) {
-        e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" onclick="showHidePipelineSection(this); return false">hide</a>)</span>')
+        e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" class="pipeline-toggle">hide</a>)</span>')
         // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
+
+        const toggle = e.querySelector('.pipeline-show-hide:last-child .pipeline-toggle');
```
basil, line 20: Prefer  `querySelectorAll` to reliably select the last `.pipeline-toggle` element, as it works regardless of the DOM structure:

```suggestion
        const toggles = e.querySelectorAll('.pipeline-show-hide .pipeline-toggle');
        const toggle = toggles[toggles.length - 1];
```

Picking the last item in the list ensures the correct element is always selected, even if `e` contains multiple `.pipeline-show-hide elements`. In contrast, `querySelector('.pipeline-show-hide:last-child .pipeline-toggle')` only works if `.pipeline-show-hide` is the last child of its parent, making it fragile if the DOM changes.

shlomomdahan: Understood, sorry for the back and forth with this change. 

I appreciate your explanation and will note it moving forward.

src/main/resources/org/jenkinsci/plugins/workflow/job/console/NewNodeConsoleNote/script.js
```diff
@@ -14,8 +14,14 @@ Behaviour.specify("span.pipeline-new-node", 'NewNodeConsoleNote', 0, function(e)
...
     var startId = e.getAttribute('startId')
     if (startId == null || startId == nodeId) {
-        e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" onclick="showHidePipelineSection(this); return false">hide</a>)</span>')
+        e.innerHTML = e.innerHTML.replace(/.+/, '$&<span class="pipeline-show-hide"> (<a href="#" class="pipeline-toggle">hide</a>)</span>')
         // TODO automatically hide second and subsequent branches: namely, in case a node has the same parent as an earlier one
+        const toggles = e.querySelectorAll('.pipeline-show-hide .pipeline-toggle');
+        const toggle = toggles[toggles.length - 1];
+        toggle.addEventListener('click', function(event) {
+            event.preventDefault();
+            showHidePipelineSection(this);
```
basil, line 23: As in https://github.com/jenkinsci/job-config-history-plugin/pull/338#discussion_r1795056135, I think it would be more readable to use `toggle` instead of `button` here.

basil (approved):
Thanks @shlomomdahan! Approving pending one last piece of feedback. Can you please test this in ATH as in https://github.com/jenkinsci/acceptance-test-harness/pull/1772 and then I will be ready to merge/release this.

PR #479, opened 2024-10-19
Refine content and appearance of the project 'Configure' screen

janfaracik:
Relates to https://github.com/jenkinsci/jenkins/pull/9734.

**Key changes**

* The 'Build Triggers' section has been updated to feature the same content changes as core, it's now just 'Triggers' and there is now a description
* Pipeline section has been moved above 'Advanced' and a description has been added
*  'Quiet period' has been moved to 'Advanced', consistent with core

**In Jenkins 2.491**

image

**In Jenkins 2.479.1 (no regressions)**

image

### Testing done

* Works as before

### Submitter checklist


pom.xml
```diff
@@ -64,7 +64,8 @@
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
         <!-- TODO: Waiting for JENKINS-73824 and JENKINS-73835 to make it into an LTS line -->
-        <jenkins.version>2.481</jenkins.version>
+        <!-- <jenkins.version>2.481</jenkins.version> -->
+        <jenkins.version>2.481-rc35491.56cb_9df6e9cd</jenkins.version>
```
jglick, line 68: Should be in draft so long as this is using an incremental version.

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowJob/configure-entries.properties
```diff
@@ -0,0 +1 @@
+pipeline.description=Define your Pipeline using Groovy directly or pull it from source control. Check the <a target="_blank" href="./pipeline-syntax">Pipeline Syntax</a> page for guidance on writing pipelines.
```
jglick, line 1: A bit pedantic, but the `pipeline-syntax` link is actually defined in `workflow-cps` and specific to those `FlowDefinition` implementations. Can this blurb be kept in `workflow-cps`, and just shifted up in the `config.jelly`? Or can this aspect of the redesign be split out for separate consideration so we can go ahead with the rest, which certainly seems welcome?

janfaracik: Definitely :)

jglick (approved):

PR #531, opened 2025-05-02
JENKINS-74908 Stop the previous build immediately

Seros:
This change ensures that previous builds are aborted immediately as expected so new builds don't need to handle possible clutter. Possibly fixes jenkinsci/bitbucket-branch-source-plugin#768 as well.

### Testing done

I tested this in a local Jenkins instance by running the same job twice concurrently and having the first build being aborted immediately as expected.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -345,8 +345,9 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                 if (prev != null && prev.isBuilding()) {
                     Executor e = prev.getExecutor();
                     if (e != null) {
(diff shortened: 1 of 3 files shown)


jglick (changes requested):
No, `doKill` should be reserved only as a last resort. Whatever the BB problem is, it should be addressed some other way.

Seros:
I see. Would `doTerm` be a better way to achieve what is wanted as it's less hard and gives the option to forcibly kill the entire build?

jglick:
`doTerm` is less bad but still not wanted here,and would not necessarily address the problem in the design of the BB plugin since that does not guarantee immediate termination either. The build number used for the last commit status update could be remembered in the plugin (or looked up from the existing commit status via API), so that the plugin would decline to post a status that would clobber an existing status based on a newer build number. Ultimately the user workflow is problematic here since if there are multiple builds associated with the same commit, it is unclear what really should be reported; it is more transparent to just push an amending commit if you want a new build.

PR #544, opened 2025-06-27
Avoid reëntrant call to `getExecution` from `isInProgress`

jglick:
https://github.com/jenkinsci/workflow-cps-plugin/pull/1047#issuecomment-3010759190


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -814,6 +826,9 @@ public boolean hasntStartedYet() {
     @Override public boolean isInProgress() {
```
Vlatombe, line 826: I guess this is called by declarative somewhere?

dwnusbaum (approved):

PR #547, opened 2025-07-01
Split `WorkflowRun.GraphL` in two and use new `GraphListener.ordinal` method to improve listener ordering

dwnusbaum:
See https://github.com/jenkinsci/workflow-cps-plugin/pull/1052 and https://github.com/jenkinsci/workflow-api-plugin/pull/410.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -328,6 +328,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
             synchronized (getMetadataGuard()) {  // Technically safe but it makes FindBugs happy
                 FlowExecutionList.get().register(owner);
                 newExecution.addListener(new GraphL());
+                newExecution.addListener(new TimingListener());
                 newExecution.addListener(new NodePrintListener());
                 completed = Boolean.FALSE;
                 executionLoaded = true;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -1085,12 +1087,16 @@ private final class FailOnLoadListener implements GraphListener {
         }
     }
 
-    private final class GraphL implements GraphListener {
+    private final class TimingListener implements GraphListener {
         @Override public void onNewHead(FlowNode node) {
             if (node.getPersistentAction(TimingAction.class) == null) {
                 node.addAction(new TimingAction());
             }
```
dwnusbaum, line 1094: I tried to write a test for the new logic and noticed that this seems to be largely redundant because of [`FlowHead.setNewHead`](https://github.com/jenkinsci/workflow-cps-plugin/blob/e20cabaa281548541b4f1398592937048e445eab/plugin/src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java#L138). I think the `if` was added in https://github.com/jenkinsci/workflow-job-plugin/pull/75 to prevent duplicate actions after https://github.com/jenkinsci/workflow-cps-plugin/pull/188 started adding `TimingAction` in `FlowHead` (merged same day, but the former depends on the latter). Maybe though the whole block should be deleted here instead. I'm not sure if there's a case where it does anything.

PR #548, opened 2025-07-01
Stop adding `TimingAction` in `WorkflowRun$GraphL` and let `workflow-cps` handle it instead

dwnusbaum:
Requires https://github.com/jenkinsci/workflow-cps-plugin/pull/1052 to be release first, and then when this is released, users need to make sure that `workflow-cps` is updated either before or at the same time that they update to this PR, otherwise `FlowStartNode` will end up without a `TimingAction`.

While trying to write tests for https://github.com/jenkinsci/workflow-cps-plugin/pull/1052 and https://github.com/jenkinsci/workflow-job-plugin/pull/547, I realized that the way that `WorkflowRun$GraphL` adds `TimingAction` is almost totally redundant due to `FlowHead.setNewHead` already adding `TimingAction`. The `TimingAction` addition logic was added to `FlowHead.setNewHead` in https://github.com/jenkinsci/workflow-cps-plugin/pull/188, and then the `if` was added to the logic here in https://github.com/jenkinsci/workflow-job-plugin/pull/75 when picking up the `workflow-cps` PR in tests to avoid adding a second `TimingAction`.

The logic in `workflow-job` is almost totally unused now, since `setNewHead` is responsible for calling `notifyNewHead`, which is what notifies `GraphListener`s, so it already handles almost all nodes. The only case that still matters is `FlowStartNode`, because `newStartNode` is separate from `setNewHead`, and `newStartNode` doesn't add `TimingAction`. https://github.com/jenkinsci/workflow-cps-plugin/pull/1052 now handles `FlowStartNode` directly as of https://github.com/jenkinsci/workflow-cps-plugin/pull/1052/commits/42419079a10571f28e5387c546f00c7aa8c9892b.

### Testing done

I tested this PR against the new automated tests in https://github.com/jenkinsci/workflow-cps-plugin/pull/1052, to ensure that with this PR and without the changes to `FlowHead.newStartNode` in that PR, the new test `CpsFlowExecutionTest.timingActionAlwaysAdded` fails. (I could duplicate the same test here, but it seems clearest to test over in `workflow-cps` since that plugin is the one adding `TimingAction`.)

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -95,7 +95,6 @@ import jenkins.scm.RunWithSCM;
 import jenkins.util.SystemProperties;
 import jenkins.util.Timer;
 import org.jenkinsci.plugins.workflow.FilePathUtils;
-import org.jenkinsci.plugins.workflow.actions.TimingAction;
 import org.jenkinsci.plugins.workflow.flow.BlockableResume;
 import org.jenkinsci.plugins.workflow.flow.DurabilityHintProvider;
 import org.jenkinsci.plugins.workflow.flow.FlowCopier;


dwnusbaum:
Ok, it looks like right now this logic is still needed for `FlowStartNode`, but I think that's the only case.

PR #561, opened 2025-09-23
Clearer fix of `isInProgress` recursion

jglick:
The fix in #544 was pretty weird and I found that it does not cover cases where a `RunAction2` calls (for example) `isLogUpdated`, which seems reasonable to do. I am reverting 430daf3b21dc9911f6c65a2aefa3f83d80e2b49a (from #96) since, whatever its original intent (not entirely clear; perhaps obsoleted by https://github.com/jenkinsci/workflow-cps-plugin/pull/675?), it is not an acceptable approach: you would not expect a `boolean` getter (an override, at that) to have major side effects such as loading the program.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>5.17</version>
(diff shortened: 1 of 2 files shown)


jglick:
Keeping in draft while I do a bit of testing.

PR #274, opened 2022-06-29
Modernize icons

DuMaM:
This PR add new svg icon from set used across whole Jenkins -> https://ionic.io/ionicons
This follow with more modernize Jenkins look and enable theming.
More here: https://weekly.ci.jenkins.io/design-library/Symbols/

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

Other connected PR's:
https://github.com/jenkinsci/workflow-cps-plugin/pull/559
Note: This require latest LTS as a base: 2.346.1

Project Status:
FireShot Capture 009 - qmk  Jenkins  - 192 168 0 31

Build Status:
FireShot Capture 010 - qmk #178  Jenkins  - 192 168 0 31

Project Side Panel:
FireShot Capture 011 - autoupdate  Jenkins  - 192 168 0 31

Build Side Panel:
FireShot Capture 012 - autoupdate #193  Jenkins  - 192 168 0 31

PR #746, opened 2026-03-26
Remove edit build information sidepanel item - moved to core

timja:
Noticed in https://github.com/jenkinsci/jenkins/pull/26525#issuecomment-4130346934

Easiest looks to be to depend on weekly, no other changes in flight that this will effect.

### Testing done

Only the new edit build information is there now:

image

(The old one had a different icon)

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,8 +64,9 @@
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
         <!-- https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ -->
-        <jenkins.baseline>2.504</jenkins.baseline>
-        <jenkins.version>${jenkins.baseline}.3</jenkins.version>
+        <jenkins.baseline>2.541</jenkins.baseline>
+        <!-- TOOD change back to jenkins.baseline after next LTS is released -->
+        <jenkins.version>2.556</jenkins.version>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
         <hpi.compatibleSinceVersion>2.26</hpi.compatibleSinceVersion>
@@ -78,7 +79,7 @@
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
                 <artifactId>bom-${jenkins.baseline}.x</artifactId>
-                <version>5353.v3dec76e40169</version>
+                <version>6166.va_a_8b_5eda_8ef5</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
--- a/src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
@@ -32,16 +32,6 @@
 
             <j:choose>
                 <j:when test="${newBuildPage}">
-                    <!-- Build Information -->
-                    <j:choose>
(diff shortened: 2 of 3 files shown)


janfaracik (approved):

src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowRun/sidepanel.jelly
```diff
@@ -32,16 +32,6 @@
 
             <j:choose>
                 <j:when test="${newBuildPage}">
-                    <!-- Build Information -->
-                    <j:choose>
-                        <j:when test="${h.hasPermission(it,it.UPDATE)}">
-                            <l:task icon="symbol-create-outline plugin-ionicons-api" href="${buildUrl.baseUrl}/configure" title="${%Edit Build Information}"/>
-                        </j:when>
-                        <j:otherwise>
-                            <l:task icon="symbol-view plugin-workflow-job" href="${buildUrl.baseUrl}/configure" title="${%View Build Information}"/>
```
jglick, line 41: Did this leave behind an unused image? https://github.com/jenkinsci/workflow-job-plugin/pull/274 but that seems to be slightly different. I am confused by the `plugin-workflow-job` here since I can find no `view.svg` here.

timja: It must've been using https://github.com/jenkinsci/jenkins/blob/master/war/src/main/resources/images/symbols/view.svg

I guess its visible from the plugin classloader when it loads the resource?

https://github.com/jenkinsci/jenkins/blob/0e852bdbce6d532f4793556ef96093f0a6633e28/core/src/main/java/org/jenkins/ui/symbol/Symbol.java#L97

pom.xml
```diff
@@ -64,8 +64,9 @@
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
         <!-- https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ -->
-        <jenkins.baseline>2.504</jenkins.baseline>
-        <jenkins.version>${jenkins.baseline}.3</jenkins.version>
+        <jenkins.baseline>2.541</jenkins.baseline>
+        <!-- TOOD change back to jenkins.baseline after next LTS is released -->
+        <jenkins.version>2.556</jenkins.version>
```
jglick, line 69: @timja what is the significance of 2.556 in this context? I am not seeing anything in https://github.com/jenkinsci/jenkins/releases/tag/jenkins-2.556 that would explain why this could not be targeted to 2.555. https://github.com/jenkinsci/jenkins/pull/26525 was merged in 2.557. Am I missing something?

timja: Its https://github.com/jenkinsci/jenkins/pull/11204

If it was set to 2.555 then there would be no edit build information button I think.

timja: It was noticed after that was shipped, see https://github.com/jenkinsci/jenkins/pull/26525#issuecomment-4130346934

Letting core handle this meant the link only showed up once not twice.

jglick: OK, just trying to figure out how to resolve a novel (to me) problem: I want to deliver some fixes to 2.555.x users, but I need at least 2.516.x and this PR jumped across that. So if I create a backport branch based on the last release prior to this PR, I would have to bump the core dep in the branch, precluding the possibility of offering something older than that later on the branch. I guess that is so old no one cares any more.

PR #542, opened 2025-06-16
Simplified locking in `getExecution`

jglick:
Amends #97. A deadlock was reported

<details>
<summary>Redacted dump</summary>

```
…"jenkins.util.Timer [#4]" id=63 (0x3f) state=BLOCKED cpu=86%
    - waiting to lock <0x7e3b640a> (a org.jenkinsci.plugins.workflow.job.WorkflowRun)
      owned by "Jetty (winstone)-397" id=397 (0x18d)
    at PluginClassLoader for workflow-job//org.jenkinsci.plugins.workflow.job.WorkflowRun.getExecution(WorkflowRun.java:709)
    at PluginClassLoader for workflow-job//org.jenkinsci.plugins.workflow.job.WorkflowRun.onLoad(WorkflowRun.java:591)
    at hudson.model.RunMap.retrieve(RunMap.java:273)
    at hudson.model.RunMap.retrieve(RunMap.java:65)
    at jenkins.model.lazy.AbstractLazyLoadRunMap.load(AbstractLazyLoadRunMap.java:703)
    at jenkins.model.lazy.AbstractLazyLoadRunMap.load(AbstractLazyLoadRunMap.java:685)
    at jenkins.model.lazy.AbstractLazyLoadRunMap.getByNumber(AbstractLazyLoadRunMap.java:579)
    at jenkins.model.lazy.LazyBuildMixIn.getBuildByNumber(LazyBuildMixIn.java:239)
    at PluginClassLoader for workflow-job//org.jenkinsci.plugins.workflow.job.WorkflowJob.getBuildByNumber(WorkflowJob.java:233)
    at PluginClassLoader for workflow-job//org.jenkinsci.plugins.workflow.job.WorkflowJob.getBuildByNumber(WorkflowJob.java:104)
    at hudson.model.Run.fromExternalizableId(Run.java:2472)
    at …

"Jetty (winstone)-397" id=397 (0x18d) state=BLOCKED cpu=84%
    - waiting to lock <0x157e1ed5> (a java.lang.Object)
      owned by "jenkins.util.Timer [#4]" id=63 (0x3f)
    at PluginClassLoader for workflow-job//org.jenkinsci.plugins.workflow.job.WorkflowRun.finish(WorkflowRun.java:644)
    at PluginClassLoader for workflow-job//org.jenkinsci.plugins.workflow.job.WorkflowRun$GraphL.onNewHead(WorkflowRun.java:1076)
[...]

PR #747, opened 2026-03-30
Fix a deadlock between `metadataGuard` and `this` (WorkflowRun) caused by a lock ordering violation in `WorkflowRun.GraphL.onNewHead()`

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### Submitter checklist


jglick:
From an extremely brief glance, might solve the issue, but looks risky. https://github.com/jenkinsci/workflow-job-plugin/pull/542#issuecomment-3189318992 I am more inclined to make https://github.com/jenkinsci/workflow-cps-plugin/pull/1008 the default (or simply delete the placeholder hack altogether).

jglick:
Might be right, but I am hoping the need for this will be obviated by just suppressing the code path which is known to introduce the deadlock.

PR #753, opened 2026-04-23
Impose a limit on how old builds may be that get resumed

jglick:
There is no reason in principle why a Pipeline build might not legitimately run for months, for example in some sort of release orchestration capacity. In practice this seems unlikely, and quite old builds not marked as complete are generally in some sort of inconsistent state. Might address unreproducible issues such as https://github.com/jenkinsci/workflow-durable-task-step-plugin/issues/604#issuecomment-3646894704 or https://github.com/jenkinsci/workflow-job-plugin/issues/730#issuecomment-3647035295. Defaulting to a 30-day grace period but this can be customized via system property.

[CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-57844)


jglick:
Breaks various `@LocalData` tests such as `FlowExecutionListTest.resumeStepExecutionsWithCorruptFlowGraphWithCycle`. Maybe the function should be disabled in tests other than the one introduced `WorkflowRunTest`.

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -629,6 +640,8 @@ public boolean hasAllowKill() {
         }
     }
 
+    static Duration MAX_RESUMPTION_AGE = Main.isUnitTest ? null : SystemProperties.getDuration(WorkflowRun.class.getName() + ".maxResumptionAge", Duration.ofDays(30));
```
rsandell, line 643: So no need to be able to turn this off via systemproperty? YAGNI I guess :) or set the duration to the heat death of the universe.

rsandell (approved):

src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: > the heat death of the universe

Not sure if that exceeds 2³¹ days but yeah.

